PR #35, opened 2016-02-22
github token and namespace support in PR bot

resmo:
I implemented github private token support, useful if you enabled 2fa
(you should...). You can then just pass --ghtoken <token> to the cli.
Find more about gh token here
https://help.github.com/articles/creating-an-access-token-for-command-line-use/

Further I implemented support for "namespaces":

e.g. cloud/cloudstack/: resmo

This means I would also be added as maintainer for any PR in this namespace.

### DIFF
--- a/MAINTAINERS-EXTRAS.txt
+++ b/MAINTAINERS-EXTRAS.txt
@@ -37,27 +37,10 @@ cloud/centurylink/clc_modify_server.py: clc-runner
 cloud/centurylink/clc_publicip.py: clc-runner
 cloud/centurylink/clc_server.py: clc-runner
 cloud/centurylink/clc_server_snapshot.py: clc-runner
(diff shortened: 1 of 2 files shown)


gregdek:
THANK GOD

gregdek:
That'll teach me to merge before testing. :)

The issue appears to be that the new handle_url code doesn't appear to be returning properly. So r.links, which expects to be this:

`{'last': {'url': 'https://api.github.com/repositories/24474138/pulls?state=open&page=10', 'rel': 'last'}, 'next': {'url': 'https://api.github.com/repositories/24474138/pulls?state=open&page=2', 'rel': 'next'}}`

...is instead just {}.

It may also be that the new code doesn't handle user/pass, which is still what we're using.

PR #50, opened 2016-03-23
workaround for unknown mergeable state

resmo:
we often get a unknown state from the api for mergeable. I don't see why, but this is a workaround.

### DIFF
--- a/triage.py
+++ b/triage.py
@@ -20,6 +20,7 @@ from __future__ import print_function
 import argparse
 import os
 import sys
+import time
 from datetime import datetime
 
 from github import Github
@@ -135,6 +136,11 @@ class PullRequest:
 
     def is_mergeable(self):
         """Return True if PR is mergeable"""
+        while self.instance.mergeable_state == "unknown":
+            print("Mergeable state is unknown, trying again...")
+            time.sleep(1)
+            self.instance = self.repo.get_pull(self.pr_number)
+            time.sleep(1)
         return self.instance.mergeable_state != "dirty"
 
     def get_base_ref(self):


gregdek:
Hunh. I've never seen "unknown" as a state. How long does it generally persist?

resmo:
My experience is, on the second api call (which is the workaround) I get a real state from the api (clean, dirty). 

So it seems that the api does not wait for the mergeable check which obviously runs in the background.

Sometimes you can also see it in the web ui (IMHO) that it takes some time to find out if it is mergeable or not

gregdek:
OK. I'll test it out with my next triage run.

resmo:
It prints a "mergeable state unknown, trying again" so you can see it when it happens.

gregdek:
Looks code. Funny corner case: when you accidentally run it against a merged PR, it doesn't like that. But that's nothing we should worry about. :)

gregdek:
Ah, spoke too soon:

PR #3307: WIP: Making unarchive idempotent
Created at 2016-03-23 13:46:42
Updated at 2016-03-23 16:28:15
Debug: PR is not mergeable
Traceback (most recent call last):
  File "./triage.py", line 684, in <module>
    main()
  File "./triage.py", line 681, in main
    triage.run()
  File "./triage.py", line 629, in run
    self.process()
  File "./triage.py", line 572, in process
    self.add_desired_label_by_build_state()
  File "./triage.py", line 308, in add_desired_label_by_build_state
    build_status = self.pull_request.get_build_status()
  File "./triage.py", line 127, in get_build_status
    return build_statuses[0]
  File "/Library/Python/2.7/site-packages/github/PaginatedList.py", line 40, in __getitem__
    return self.__elements[index]
IndexError: list index out of range

PR #61, opened 2016-03-29
triage: commit after last comment scenario

resmo:
/cc @gregdek

### DIFF
--- a/triage.py
+++ b/triage.py
@@ -379,6 +379,8 @@ class Triage:
 
         self.debug(msg="--- START Processing Comments:")
 
+        last_commit_date = self.pull_request.get_build_status().updated_at
+
         for comment in comments:
 
             # Is the last useful comment from a bot user?  Then we've got a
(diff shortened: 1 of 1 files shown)


gregdek:
OK, so for clarity: is this for the "shipit but there have been edits so don't shipit" case?

And if so, should we be messaging this somehow?

resmo:
@gregdek exaclty, That would make sense. 

E.g. Something like: 

"Thanks @{{ submitter }}, we see you did some work after the maintainers review. Therefore we put the PR back in community review. To the current maintainers, @{{ maintainer|join(', @') }} please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate."

gregdek:
OK. If you can add that logic to this PR, I'll merge it. Thanks!

resmo:
will rebase

gregdek:
Still a little confused here, so bear with me.

1. I still think we need separate boilerplate content for this, noting that we're putting it into *_review because the PR was changed. 

2. Would this apply to a PR that's already in shipit? Looking at #L333, wouldn't it just skip this entire path because the PR is already labeled shipit? Or is this PR not intended to handle that case?

resmo:
1. Suggestions are very welcome, I am sure you find better words than I do.
2. We skip the maintainers L603 for shipits, but still handle the comments on L606.

gregdek:
I think this still needs to point to the new boilerplate file, right? If this PR can use the messaging in comment 3, I think we're ready to go.

resmo:
closing since this is not relevant anymore IMHO

PR #65, opened 2016-03-30
triage: implement new owner_pr process

### LINKED ISSUE #54: Change "owner_pr" process
Per discussion on mailing list:

* do away with owner_pr tag, as it's superfluous

* new owner_pr boilerplate that says: "This module is going into community review. We notice that you are maintainer of this module, so when you think it's ready, please comment "shipit" and we will consider it for merging."

* change the current "owner_pr" behavior to put an owner_pr into community_review instead of shipit.

resmo:
Closes #54

@gregdek: FYI, found 2 PR e1370 and e1931 where I quick removed owner_pr and then ran the bot, which gave the result wanted.

actually I noticed the owner_pr is not even needed... we can get rid of if you like, I would keep it.


gregdek:
let's leave the owner_pr tag for now.

gregdek:
Thanks @resmo.

PR #67, opened 2016-03-30
triage: implement WIP

### LINKED ISSUE #53: New pr bot state: "WIP"
The goal of the "WIP" (work in progress) state is to tell the bot to just leave it alone for now. Proposal:

* Any PR with "WIP or [WIP]" at the start of the title is labeled "WIP"
* Any PR labeled "WIP" but does not have "WIP" in the title has label removed
* Boilerplate telling people "this PR is a work-in-progress. @submitter, when you're ready to have this PR reviewed, please remove WIP from the title of the PR."
* Any PR in WIP is otherwise ignored completely by the bot.

resmo:
closes #53 

NOTE: there is no label "work_in_progress" yet defined in the repos!

### DIFF
--- a/triage.py
+++ b/triage.py
@@ -160,6 +160,12 @@ class PullRequest:
             time.sleep(1)
         return self.instance.mergeable_state != "dirty"
 
+    def is_a_wip(self):
+        """Return True if PR start with [WIP] in title"""
+        return (self.instance.title.startswith("[WIP]")
+                or self.instance.title.startswith("WIP:")
+                or self.instance.title.startswith("WIP "))
+
     def get_base_ref(self):
         """Returns base ref of PR"""
         return self.instance.base.ref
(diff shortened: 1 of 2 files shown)


triage.py
```diff
@@ -160,6 +160,10 @@ def is_mergeable(self):
             time.sleep(1)
         return self.instance.mergeable_state != "dirty"
 
+    def is_a_wip(self):
+        """Return True if PR has WIP in title"""
+        return "WIP" in self.instance.title or "[WIP]" in self.instance.title
```
dagwieers: The exact implementation in Gitlab is that the title should *start* with "WIP: " or "[WIP] ". Although it seems with newer Gitlab releases they only hint into using "WIP: ". So this could be reduced to:

```python
return self.instance.title.startswith('WIP: ') or self.instance.title.startswith('[WIP] ')
```
PS Or you could drop the trailing whitespace if you want to be less pedantic than me :)

resmo: @dagwieers thanks for the input, startswith makes more sense here.

gregdek:
Tested on e/1962, looks good. Added labels to repos and merging. Thanks @resmo!

PR #68, opened 2016-03-31
triage: fix backport must be evaluated after maintainers

resmo:
/cc @gregdek 

otherwise this happens... c3353


gregdek:
Thx @resmo!

PR #69, opened 2016-03-31
triage: fix case in submitter name / maintainer name

resmo:
/cc @gregdek 
missed the case insensitive condition here


gregdek:
Thx @resmo!

Issue #44, opened 2016-03-03
Add a "close_me" state


gregdek:
Sometimes a PR just isn't going to be accepted; maybe it's a duplicate, for instance. In that case, we should provide a mechanism for the maintainer of the module to request closure of the module.

gregdek:
Should be a simple thing: 

1. Add close_me as an option to all the boilerplate;
2. Detection of close_me by a maintainer results in immediate closure with new boilerplate.

jctanner:
Need to re-add for triagev3.

PR #84, opened 2016-04-16
implement close_me

resmo:
Implements #44 

/cc @gregdek 

Note: template text should be revisioned and a new label close_me must be added.


gregdek:
/me thinks.

Either we want this action to be fully automatic, or semi-automatic.

If we want it to be fully automatic, we would simply close the PR. No label needed.

If we want it to be semi-automatic, which means we need to review occasionally, then I would prefer to use the already existing "pending_action" label, which we review weekly anyway. 

So I would say: let's change this to use the pending_action label.

resmo:
@gregdek ready for review

gregdek:
I like it. Thanks @resmo!

PR #92, opened 2016-05-03
fix repeated maintainer unknown reporting

resmo:
this fixes repeated maintainer unknown reporting by adding a label pending.
/cc @gregdek


gregdek:
Thanks @resmo!

PR #103, opened 2016-06-17
add gce label for google modules

resmo:
/cc @gregdek


gregdek:
So simple. Thanks @resmo :)

PR #141, opened 2016-08-30
cloud/smartos and network/illumos maintainer

xen0l:
authored few modules, which I would like to maintain and monitor for updates. I can also help with illumos/SmartOS subject matter expert if there is a desire.


xen0l:
@gundalow I found out that you already did PR. Could you add cloud/smartos category there, too?

gundalow:
Thank you for the offer :)

PR #182, opened 2016-10-13
Add Tom M as maintainer for new GCE MIG module

ryansb:


gundalow:
Just an idea, we could change this to:

```
cloud/google/: supertom
cloud/google/gc_storage.py: bennojoy supertom
```
And delete the other `cloud/google/...` lines
So Tom M gets notified on everything, and bennojoy also gets notified on changes to `gc_storage`.

This means we don't need to keep adding people

ryansb:
Done

PR #207, opened 2016-11-14
Remove dazworrall from cloudstack

DazWorrall:
Remove myself from the cloudstack maintainers list, I don't work with it any more unfortunately.


gundalow:
Thanks for you help with everything in the past. 
@remo FYI

PR #211, opened 2016-11-16
Add labels for the different cloud providers

emonty:
The clouds are all maintained by different teams - adding a label makes
searching for revelant PRs super easy.

I would have just had it be a direct mapping, but there are no rackspace or linode tags, and the amazon/aws and google/gce mapping meant just blowing it out into a set of elifs seemed cleanest


jctanner:
@emonty _prbot.py is no longer an active file. Changes should be made in triage.py

PR #227, opened 2016-12-07
Add jhawkesworth as maintainer of win_say

jhawkesworth:
Perhaps not the best time for this PR, given module reorganisation, but want to add myself as maintainer of this module for future reference.


gundalow:
Merged, though I'm not sure if we have the bot running at the moment, we will do again shortly. So getting extra ownership details in is good.

PR #245, opened 2017-01-05
Notify users when the PR contains travis

jctanner:


templates/travis_notify.j2
```diff
@@ -0,0 +1,4 @@
+@{{ submitter }} This PR contains references to travis-ci.org. Please rebase your branch and push a new commit. [rebase instructions](http://docs.ansible.com/ansible/dev_guide/developing_rebasing.html)
```
mattclay: Suggested revision for this sentence:

```
@{{ submitter }} This PR was tested by travis-ci.org, which is no longer used.
Please [rebase your branch](http://docs.ansible.com/ansible/dev_guide/developing_rebasing.html) to trigger running of current tests.
```

mattclay: Is having a custom message for Travis going to cause a problem when you add one for needs_rebase? Won't you end up with both messages added to the comments?

PR #261, opened 2017-01-16
Updates maintainers to add F5 devs for all F5 modules

caphrim007:
This patch updates the F5 modules so that F5 devs are the maintainers
for all of them and so that the community contributors are no longer
on the hook for their support


caphrim007:
@gundalow as requested

PR #272, opened 2017-01-18
Remove duplicate lines

jasperla:


gundalow:
Good spot, some of those will have been when we merged `-core` and `-extra` lists

PR #328, opened 2017-02-06
Add ovirt label & classifier for ovirt PRs

ryansb:
The oVirt team wants to use Ansible GH instead of their issue tracker for the modules, and have requested that we have an oVirt label so they're able to find & tag issues that apply to them.


jctanner:
The bot already adds labels by module topic and subtopic -if- the relevant label exists. In this case, we only needed to create the "ovirt" label to make it work.

https://github.com/ansible/ansible/pull/20310

PR #405, opened 2017-03-03
Add a fetcher+parser for shippable test results

jctanner:


lib/utils/shippable_api.py
```diff
@@ -79,3 +85,87 @@ def get_updated_since(self, since):
...
+                    results.append(cdata)
+
+                else:
+                    # sometimes it is xml ...
+                    root = None
+                    try:
+                        root = objectify.fromstring(contents)
+                    except ValueError:
+                        # sometimes it has non-serializable unicode
+                        contents = contents.encode('ascii', 'ignore')
```
mattclay: What data did you encounter that can't be handled as utf-8?

PR #436, opened 2017-03-22
Removing maintainer

caphrim007:
mhite has a long and storied past with the F5 modules, but in his
day-to-day work he deals with them much less and caphrim007, who
works at F5, has volunteered to take over the maintenance of all
the modules. Removing him so that he's not notified of future
issues/prs


gundalow:
Thanks for the update.

PR #491, opened 2017-04-19
add missing s3_sync owner

tedder:
That's why I wasn't notified.


gundalow:
@tedder Thanks for this, we occasionally don't get the list updated when we add new modules, though we do a big sweep every so often. This change will take effect next time the bot is restarted.

PR #543, opened 2017-05-26
Allow to flag variables as mandatory

pilou-:
ansibot doesn't work when Github credentials are missing, this pull-request checks if credentials are set.

### DIFF
--- a/lib/constants.py
+++ b/lib/constants.py
@@ -63,8 +63,8 @@ def shell_expand(path, expand_relative_paths=False):
     return path
 
 
-def get_config(p, section, key, env_var, default,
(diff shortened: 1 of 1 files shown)


jctanner:
@pilou- I like the patch, but requiring a password was not the initial intent. I would like to allow the user to use a token (I will move ansibot to that at some point).

pilou-:
Ok. Besides I saw that `constants.py` was fetched form ansible repository, importing would be better. That's why I am working on splitting constants.py (keeping only constants there and moving methods elsewhere): https://github.com/pilou-/ansible/commit/3212d585b51a1dfc386481ed7354f3e2f3c244ca.

PR #547, opened 2017-05-26
Don't fail when rate limit response doesn't contain requested information, improve consistency

pilou-:
* don't assume that rate limit response contain usable information
* [some code assumes](https://github.com/ansible/ansibullbot/blob/6fbf34c8c815b655b547bfa0e402cddec97df0b4/lib/decorators/github.py#L41) that `get_rate_limit` might return a `False` expression: improve consistency


jctanner:
Thanks @pilou- !

PR #569, opened 2017-06-09
update win_reboot doc owner

nitzmahone:


gundalow:
What about `windows/win_reboot.ps1`?
Should we just use `windows/win_reboot` (without extension) to match both `.py` & `.ps1`?

nitzmahone:
It's an action plugin- this file seems to be rooted in `modules`, so I'm guessing that won't work...

PR #631, opened 2017-07-08
Add maintainer for zfs_facts.py and zpool_facts.py

xen0l:


jctanner:
@xen0l with the migration to https://github.com/ansible/ansible/blob/devel/.github/BOTMETA.yml module authors will be automatically marked as maintainers, so we no longer need this PR.

Thanks

PR #632, opened 2017-07-09
Added win_wakeonlan, cleaned up my other entries.

dagwieers:


jctanner:
@dagwieers we've migrated to https://github.com/ansible/ansible/blob/devel/.github/BOTMETA.yml and you are marked as the maintainer for those modules.

dagwieers:
@jctanner Great, thanks !

PR #641, opened 2017-07-15
Removing customer ownership

caphrim007:
Removing customer ownership. F5 maintains these modules now


jctanner:
@caphrim007 this file is just test data to validate the parser. The production data is in https://github.com/ansible/ansible/blob/devel/.github/BOTMETA.yml

PR #658, opened 2017-07-18
Fetch only specified pull-requests when 'pr' switch used

pilou-:
This pull-request allows to fetch only pull-requests specified with `pr` switch.


jctanner:
@pilou- awesome! I wanted to do this but hadn't gotten to it yet. I will test and hopefully merge in the morning.

Issue #460, opened 2017-03-31
Re-run CI on `stale_ci` PRs waiting to be merged.


mattclay:
If a PR is waiting to be merged, but is blocked **only** due to `stale_ci`, CI should be re-run.

Depends on #459

jctanner:
After seeing a burst of new PRs from the filament/lightbulb PRs, I have the impression that we don't have enough shippable resources to handle this.

pilou-:
The CI could be re-run for pull-requests:
* blocked only due to `stale_ci`
* and if last shippable result for this pull-request is `success`
* and with at least one `shipit`

Here is an implementation: #664.

mattclay:
This feature won't produce bursts of CI jobs since it will only occur when PRs have been approved for merge and are stale.

gundalow:
@mattclay Is this still an issue?

mattclay:
I haven't watched recent bot merge behavior closely enough to notice if this is implemented or not.

Issue #459, opened 2017-03-31
Block auto-merge if PR is labeled `stale_ci`.


mattclay:
The bot should not auto-merge PRs if they have the `stale_ci` label applied.

Depends on #458

jctanner:
This should be done already.

PR #664, opened 2017-07-24
re-run CI on stalled pull-request with 'shipit'

pilou-:
This pull-request allow to automatically re-run CI on pull-request labeled with both `shipit` and `stale_ci`.

This will facilitate contributors and maintainers workflows when a pull-request is labeled with `stale_ci`, avoiding submitter:

1. to rebase the branch pull-request in order to trigger the CI and to remove the `stale_ci` label;
2. to ask maintainer who had already approved a pull-request using the `shipit` command to use again the `shipit` command (because `shipit` counters are reset after a rebase).

I was not able to test (on a custom project) the HTTP request to `projects/{projectId}/newBuild`, this Shippable API [isn't usable](https://github.com/Shippable/support/issues/3628#issuecomment-305091311) using a free Shippable account.

Related: #459 #460

### DIFF
--- a/ansibullbot/utils/shippable_api.py
+++ b/ansibullbot/utils/shippable_api.py
@@ -141,30 +141,8 @@ class ShippableRuns(object):
         resp = None
         if not os.path.isfile(cfile) or not jdata or not usecache:
 
-            headers = dict(
-                Authorization='apiToken %s' % C.DEFAULT_SHIPPABLE_TOKEN
-            )
-
-            resp = None
-            success = False
-            retries = 0
-            while not success and retries < 2:
-                logging.debug('%s' % url)
-
-                try:
-                    resp = requests.get(url, headers=headers)
-                except requests.exceptions.ConnectionError:
-                    time.sleep(2)
-                    continue
-
-                if resp.status_code not in [200, 302, 400]:
-                    logging.error('RC: %s' % (resp.status_code))
-                    retries += 1
-                    time.sleep(2)
-                    continue
-                success = True
-
-            if not success:
+            resp = self.fetch(url)
+            if not resp:
                 return None
 
             if resp.status_code != 400:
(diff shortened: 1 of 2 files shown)


jctanner:
@pilou- Thanks for working on this (saves me a lot of time). 

Please move the logic to determine is a rebuild is necessary to a plugin under triagers/plugins similar to some of the other ones you'll find in there and have the function called at the end of all other facts being called in AnsibleTriage.process().  We will also want to extend this to allow maintainers or submitters to retrigger their own CI with a bot command (at some point).

I would also like the rebuild to be carried out in apply_actions(), so a "ci_rebuild" key or something similar should in the EMPTY_ACTIONS dict which is then handled by apply_actions.

ansibullbot/utils/shippable_api.py
```diff
@@ -350,3 +323,60 @@ def get_test_results(self, run_id, usecache=False, filter_paths=[]):
...
+        run_url = "%s&runNumbers=%s" % (ANSIBLE_RUNS_URL, run_number)
+        response = self.fetch(run_url)
+        if not response:
+            raise Exception("Unable to fetch %r" % run_url)
+        self.check_response(response)
+
+        run_id = response.json()[0]['id']
+
+        newbuild_url = "%s/projects/%s/newBuild" % (SHIPPABLE_URL, ANSIBLE_PROJECT_ID)
+        response = self.fetch(newbuild_url, verb='post', data=json.dumps({'runId':run_id}))
```
mattclay: We need some kind of logic to protect against repeated rebuilds of PRs. Since this depends on the build status reported by GH, it can get out of sync with what Shippable is doing. This can happen due to lost or delayed web hooks. We should probably be getting the CI status from Shippable itself, instead of relying on GH.

jctanner:
Going to merge this as-is and fix up the suggestions myself. Thanks again @pilou- !

pilou-:
Wow, you were in a hurry ! Are you working on these updates or are they planned ?

PR #675, opened 2017-07-26
Allows automerge when deletion in sanity text files

pilou-:
This behavior was added by 369232a7d1285361b4e7d2d7902b763b9767e514, don't prevent it. See #659.


jctanner:
@pilou- I would rather have special handling for the sanity file than to open the gate on all PRs with >1 files.

pilou-:
That's the case: the [`for` block](https://github.com/ansible/ansibullbot/pull/675/files#diff-0ea07bcb426627b4c5a7bd6d304ccbdcR70) has special handling for sanity files and returns `False` in all other cases.

jctanner:
@pilou- yep, i should have realized that. Thanks again!

PR #679, opened 2017-07-26
WIP: Notify namespace maintainers sooner and notify fewer maintainers

pilou-:
When there is only one module maintainer (pull request is related to an existing module) or when pull request is about a new module:
- notify namespace maintainers even if module maintainer didn't approve the pull-request using `shipit`
- notify fewer namespace maintainers (at most 3).

I would like this pull requests to be discussed with community working group:
- more or less namespace maintainers could be notified
- different number of namespace maintainers could be notified
  - for existing module with only one maintainer
  - for new module

### DIFF
--- a/ISSUE_HELP.md
+++ b/ISSUE_HELP.md
@@ -149,7 +149,7 @@ The bot adds many labels on issues and pull requests.
 
 Label | Scope | Prevent automerge | Description
 --- | --- | --- | ---
-**<a name="label-stale_ci">stale_ci</a>** | pull requests | yes | Added when the last CI result is older than one week.
+**<a name="label-stale_ci">stale_ci</a>** | pull requests | yes | Added when the last CI result is older than one week. When a pull request is closed and reopened, the CI is triggered again. In some case, the bot will automatically trigger the CI when a pull request is labeled with both [`shipit`](#label-shipit) and `stale_ci`.
 **<a name="label-stale_review">stale_review</a>** | pull requests | no | Added when submitter made some updates after a reviewer requested some changes, if the submitter updates are older than seven days and the reviewer didn't update his review.
 **<a name="label-core_review">core_review</a>** | pull requests | no | In order to be merged, these pull requests must follow the [core](#core) review workflow.
 **<a name="label-community_review">community_review</a>** | pull requests | no | In order to be merged, these pull requests must follow the [community](#community) review workflow.
--- a/ISSUE_HELP.md
+++ b/ISSUE_HELP.md
@@ -61,11 +61,19 @@ If you are not sure who the pull request is waiting on, please use the [`bot_st
 
 ### When will your pull request be merged?
 
-:information_source: `Approve` pull request status is ignored, [`shipit`](#cmd-shipit) command is used by maintainer to approve a pull request.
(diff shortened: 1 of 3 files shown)


pilou-:
cc @gregdek

pilou-:
Note for myself, check that:
- submitter is not contained in maintainers variable (or remove it) in order to avoid [this](https://github.com/ansible/ansible/pull/21542#issuecomment-318228381).
- when there isn't any other namespace maintainer

gregdek:
Here's my concerns: 

1. Right now, either a maintainer wants to be a responsible party, or they don't. There's kind of no in-between. 

2. We also have no good way of choosing "which n" maintainers get notified. The first n? The last n? A random selection of n? 

@pilou- what's your specific concern? Are you yourself receiving too many notifications?

pilou-:
My specific concern: how can we decrease number of open pull requests? how could pull requests be reviewed faster?

1. when there is only one module maintainer, I want to notify some namespace maintainers sooner (I mean: at the same time the single module maintainer is notified, not after the single module maintainer had approved) because their reviews are required (in fact there is one case when their reviews won't be required: if the single module maintainer and 3 other people approved the pull-request).

2. On the other hand, I am not sure if many namespace maintainers notified (an example [here](https://github.com/ansible/ansible/pull/27245#issuecomment-317481290)) is a good thing.

About your 2nd point: a random selection of n looks good to me.

jctanner:
1. I have no problem with people being pinged earlier, but I wouldn't count on that increasing the amount of reviews given. There's already a large set of labels that categorize the issues+pulls, so maintainers no longer have an excuse to not look at them other than being busy or absent. Notifying them won't change their availability.

2. If people don't want to be notified, they can opt out of maintainership.

mscherer:
Github notifications are kinda bad however, to the point of featuring projects like octobox ( https://github.com/octobox/octobox ) on the explore page ( https://github.com/showcases/tools-for-open-source ). 

And as for notifying less maintainers, there is psychological research that tend to lead credence to the idea (e.g. Social Proof and Bystander Effect). If we notify 10 people 10 times for 10 bugs, they will likely all think someone else will deal with it, while notifying only one or two peoples would tell them "that's up to you". It may indeed not result into better results, but it can't make things worst IMHO. 

That's also how rust-highfive work ( https://huonw.github.io/blog/2015/03/rust-infrastructure-can-be-your-infrastructure/ ), and Sarah Sharp work on the topic show this did improve things: http://sarah.thesharps.us/2016/11/17/impact-of-bots-on-github-communities/

In a ideal world also, notifying people in smaller batch would help to see who answer and who don't, and then would give us data on who is more reactive, and help find place where there is a maintenance gap.

gregdek:
The presence of "too many reviewers" is a potentially real problem. One potentially policy remedy is to cut down the max number of active reviewers to, say, 3. 

But I think the real remedy lies elsewhere. I think we're already doing plenty to encourage our community to review module PRs -- and the fact is that some of them may simply not want to do it. Which means that what we *really* need, IMO, is a better mechanism to flag modules that are not being well-maintained.

Anyway, I don't think this is a tooling issue. I think it's a process issue. We need to sort out process changes we need to make, and then make the tools fit them, not vice versa. For that reason, I'm -1 on this PR.

gundalow:
@pilou- What are your thoughts with this?

mkrizek:
Closing this as there are no modules namespaces since ansible/ansible#68966

Thanks @pilou- !

PR #705, opened 2017-08-11
Updating the NetApp team.

hulquest:
Need to update the NetApp team.  Actively recruiting a replacement.


hulquest:
Cc @lmprice

gundalow:
Hi,
Looks like you've modified the test scripts, rather than the live data.

Please update https://github.com/ansible/ansible/blob/devel/.github/BOTMETA.yml

PR #710, opened 2017-08-16
Playbooks and roles for managing Ansibullbot

samdoran:


jctanner:
@samdoran is this ready to merge?

Issue #704, opened 2017-08-10
authors line format edgecase


jctanner:
https://github.com/ansible/ansible/pull/23127

```
author: Tomas Karasek <tom.to.the.k@gmail.com>, Matt Baldwin <baldwin@stackpointcloud.com>, Thibaud Morel l'Horset <teebes@gmail.com>
```

1. make sure that is parsed correctly
2. check if the blame data can correlate those emails to github nicks

gundalow:
Do we want `validate-modules` to enforce the correct format?

pilou-:
@gundalow I don't see any reason to not enforce the correct format ([`First Last (@githubid)`](http://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-block)).

A lot of people are using `First Last <first@last.example>`: should the format be updated in order to allow to specify an email or should the email be put in the copyright block ?

bcoca:
(@githubnick) <= missing from that format, it should be required on new entries ... we need to update existing

bcoca:
also, it can be , separated string and/or yaml list

pilou-:
comma separated format is used but is not documented in [dev_guide](https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/dev_guide/developing_modules_documenting.rst#documentation-block). Should it be added ?

PR #726, opened 2017-08-22
Find GitHub ID searching for email

pilou-:
The purpose of this pull request is to avoid to add module author in `.github/BOTMETA.yml` when `DOCUMENTATION.author` contains an email address but not an GitHub ID. `.github/BOTMETA.yml` should only be used when authors and maintainers differ (and `DOCUMENTATION.author` should contain GitHub ID).

Not that because `.github/BOTMETA.yml` already contains an entry for each module where `DOCUMENTATION.author` contains an email address but not an GitHub ID, nothing will change. But some entries in `.github/BOTMETA.yml` could be removed (the ones for which module authors and module maintainers are the same and `DOCUMENTATION.author` contains email address).

Commits explanation:
- second commit: implement `get_usernames_from_filename_blame` using GraphQL (allowing to map email address and GitHub ID)
- third commit: move authors initialization later, allowing `get_module_authors` to access `emailmap`
- fourth commit: use `get_usernames_from_filename_blame` using GraphQL (when Github token is set)
- fifth commit: `get_module_authors` search for email in `author` and use `emailmap`
- six commit: populate `emailmap` in GraphQL implementation of `get_usernames_from_filename_blame`

Relates: #704

### DIFF
--- a/ansibullbot/utils/gh_gql_client.py
+++ b/ansibullbot/utils/gh_gql_client.py
@@ -286,6 +286,7 @@ class GithubGraphQLClient(object):
 
         template = self.environment.from_string(QUERY_TEMPLATE_BLAME)
         committers = defaultdict(set)
+        emailmap = {}
 
         query = template.render(OWNER=owner, REPO=repo, BRANCH=branch, PATH=filepath)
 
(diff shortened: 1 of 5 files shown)


ansibullbot/utils/gh_gql_client.py
```diff
@@ -250,6 +282,58 @@ def update_node(self, node, node_type, owner, repo):
...
+            # - GraphQL 'history' neither because 'history' is like 'git log' but without '--follow'
+            email = node['author'].get('email')
+            if email and email not in emailmap:
+                emailmap[email] = github_id
+
+        for github_id, commits in committers.items():
+            committers[github_id] = list(commits)
+        return committers, emailmap
+
+    @retry(wait=wait_random(min=1, max=2), stop=stop_after_attempt(5))
```
jctanner: GraphQL has a rate limit, similar to the rest api but with it's own limit. Should we use something more like the RateLimited decorator here instead so that the wait interval is more accurate?

pilou-: While testing I didn't reach the limit but using something like the RateLimited decorator is a good idea.

jctanner:
We can worry about the rate limiting later. Taking this now, as is.

PR #730, opened 2017-08-29
Misc fixes

jctanner:
* fix issue with sorting in the issue iterator


ansibullbot/triagers/plugins/component_matching.py
```diff
@@ -0,0 +1,224 @@
...
+            for user in cmatch['notify']:
+                if user not in cmeta['owner']:
+                    cmeta['owner'].append(user)
+
+    # if any component is core, then the whole thing is core
+    if cmeta['component_matches']:
+        core = False
+        for cmatch in cmeta['component_matches']:
+            if cmatch.get('supported_by') in ['networking', 'core']:
+                core = True
```
pilou-: Either the comment above is wrong, or a `break` statement is missing.
Note that if the comment is right, `any(cmatch.get('supported_by') in ['networking', 'core'] for cmatch in cmeta['component_matches'])` could be used.

ansibullbot/triagers/plugins/component_matching.py
```diff
@@ -0,0 +1,224 @@
...
+        cmeta['is_issue'] = False
+        cmeta['is_pullrequest'] = True
+        cmeta['component_labels'] = \
+            file_indexer.get_component_labels(
+                iw.files,
+                valid_labels=valid_labels
+            )
+
+    # who owns this? FIXME - is this even used?
+    cmeta['owner'] = 'ansible'
```
pilou-: It should be `cmeta['owner'] = ['ansible']`.

PR #732, opened 2017-09-05
take in account '--repo' switch and remove duplicated code

pilou-:
Handle `--repo` switch, remove duplicated code.


jctanner:
@pilou- I still use skip_module_repos for local testing

pilou-:
`skip_module_repos` is still handled [there](https://github.com/pilou-/ansibullbot/blob/1a4d18411e05ebca6ee017f133d736dc17e03e59/ansibullbot/triagers/ansible.py#L1519).

jctanner:
Ah yes, very true.

PR #761, opened 2017-09-11
Enable final migration for all module issues.

jctanner:


pilou-:
For #760, https://github.com/ansible/ansibullbot/blob/master/ansibullbot/triagers/ansible.py#L1530 needs to be updated too.

pilou-:
Ok, `ansibullbot/triagers/ansible.py` updated there: https://github.com/ansible/ansibullbot/pull/766/files#diff-b206352f3de977e805cb5daa6f314c53.

PR #767, opened 2017-09-12
Fix 'best match' in 'ModuleIndexer': avoid to mix maintainers

pilou-:
When:

1. a module path without an entry in `BOTMETA.yml` (for example `lib/ansible/modules/packaging/os/xbps.py`)
2. contains another unrelated entry present in `BOTMETA.yml` (for example `packaging/`)
3. then maintainers of the `BOTMETA.yml` entry are added to module maintainers

This pull request:
- add unit test
- improve match: fix this issue

### DIFF
--- a/tests/unit/utils/test_module_indexer.py
+++ b/tests/unit/utils/test_module_indexer.py
@@ -64,7 +64,7 @@ class TestModuleIndexer(TestCase):
               maintainers: csim
         """
 
-        filepath = 'packaging/apt.py'
+        filepath = 'lib/ansible/modules/packaging/apt.py'
         indexer = run(textwrap.dedent(BOTMETA), ['Hubert'], filepath)
         self.assertEqual(len(indexer.modules), 1)  # ensure only fake data are loaded
         self.assertEqual(indexer.modules[filepath]['maintainers'], ['Hubert'])
@@ -88,11 +88,11 @@ class TestModuleIndexer(TestCase):
             $modules/foo/bar: jim
         """
 
-        filepath = 'foo/bar/baz.py'
+        filepath = 'lib/ansible/modules/foo/bar/baz.py'
         indexer = run(textwrap.dedent(BOTMETA), ['bob'], filepath)
         self.assertEqual(len(indexer.modules), 1)  # ensure only fake data are loaded
         self.assertEqual(set(indexer.modules[filepath]['maintainers']), set(['bob', 'jim']))  # both
-        self.assertEqual(indexer.modules[filepath]['maintainers_key'], 'foo/bar')
+        self.assertEqual(indexer.modules[filepath]['maintainers_key'], 'lib/ansible/modules/foo/bar')
 
     def test_maintainers_dont_inherit_from_directory_maintainers(self):
         """module maintainers don't inherit from parent directory maintainers (both defined in BOTMETA)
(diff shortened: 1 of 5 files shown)


jctanner:
@pilou- I think the undocumented logic is supposed to be more like ...

```yaml
- foo/bar: jim
- foo/bar/baz.py: bob
```
jim becomes a maintainer for all things in the "bar" directory, and "baz" is maintained by jim+bob

@gregdek thoughts on this?

gregdek:
If the question is "does jctanner's logic make sense?" the answer is "I think so."

If the question is something else, I don't understand it. :)

jctanner:
@pilou- okay, it seems that we DO want the inheritance. Can you correct your patch for that?

pilou-:
@jctanner I updated the test case, the behavior you described (`foo/bar: jim` in `BOTMETA` and `author: bob` in `foo/bar/baz.py`) is working without and with this pull request. I guess something is broken but I don't see what ?

jctanner:
In "test_maintainers_inherit_dont_inherit_from_directory_maintainers", I would expect ... 

```
self.assertEqual(set(indexer.modules[filepath]['maintainers']), set(['bob']))
```

to be ...

```
self.assertEqual(set(indexer.modules[filepath]['maintainers']), set(['jim', 'bob']))
```

pilou-:
Tests updated (docstrings updated)

Using `master` branch, `test_maintainers_dont_inherit_from_directory_maintainers` fails. Do you think the test is wrong or that expected behavior is already broken ?

```
$ nosetests tests/unit/utils/test_module_indexer.py # using 2c9ee357410c97f4a1d8b64aa4cf94900da1086d
maintainers don't get mix up ... FAIL
module maintainers don't inherit from parent directory maintainers (both defined in BOTMETA) ... ok
authors defined in module metadata inherit from directory maintainers specified in BOTMETA ... ok
======================================================================
FAIL: maintainers don't get mix up
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/unit/utils/test_module_indexer.py", line 70, in test_maintainers
    self.assertEqual(indexer.modules[filepath]['maintainers'], ['Hubert'])
AssertionError: Lists differ: ['Hubert', 'csim'] != ['Hubert']
[...]
Ran 3 tests in 0.037s

FAILED (failures=1)
```

jctanner:
@pilou- it's probably broken in the bot code.

jctanner:
@pilou- I think this is how it should look ...

```diff
+    def test_maintainers_dont_inherit_from_directory_maintainers(self):
+        ...
+        BOTMETA = """
+        ---
+        macros:
+            team_green:
+              - larry
+              - cow
+            modules: lib/ansible/modules
+        files:
+            docs/:jane
+            $modules/foo/bar: jim
+            $modules/foo/bar/baz.py: bob
+        """
+
+        filepath = 'lib/ansible/modules/foo/bar/baz.py'
+        indexer = run(textwrap.dedent(BOTMETA), [], filepath)
+        self.assertEqual(len(indexer.modules), 1)  # ensure only fake data are loaded
+        self.assertEqual(set(indexer.modules[filepath]['maintainers']), set(['jim', 'bob']))  # only bob, not jim
+        self.assertEqual(indexer.modules[filepath]['maintainers_key'], 'lib/ansible/modules/foo/bar/baz.py')
```

pilou-:
@jctanner Pull request updated. I removed `ModuleIndexer.maintainers` attribute because it was used by `scripts/maintainer_v1-v2.py`: is that OK ?

I will launch the bot with `--daemonize --dry-run` for a few hours and report results here.

tests/unit/utils/test_module_indexer.py
```diff
@@ -0,0 +1,278 @@
...
+
+        for k in filepaths:
+            self.assertEqual(sorted(indexer.modules[k]['maintainers_keys']),sorted(expected[k]['maintainers_keys']))
+            self.assertEqual(sorted(indexer.modules[k]['maintainers']), sorted(expected[k]['maintainers']))
+
+    def test_authors_overridden_if_entry_in_BOTMETA(self):
+        """Check that authors are overidden if BOTMETA contains an entry.
+
+        Ensure that authors defined in 'author' field of 'DOCUMENTATION' module
+        metadata are omitted when there is a matching entry in BOTMETA.yml.
```
jctanner: authors should be included, not overridden. If an author does not wish to be a maintainer, they should be added to the "ignored' field in botmeta OR be removed as an author.

pilou-: Pull request updated. Note that currently, in this case, authors are overridden.

I much prefer the behavior you expect.

pilou-:
Pull request updated:
- new tests about `ignored` field
- `ignored` field is propagated (like `maintainers`)

PR #775, opened 2017-09-15
Store git blames in sqllite

jctanner:


ansibullbot/utils/moduletools.py
```diff
@@ -53,6 +79,17 @@ def __init__(self, maintainers=None, gh_client=None):
         self.scraper_cache = os.path.expanduser(self.scraper_cache)
         self.gws = GithubWebScraper(cachedir=self.scraper_cache)
         self.gqlc = gh_client
+        self.files = []
+
+        # sqlalchemy
+        unc = 'sqlite:///'
+        unc += os.path.expanduser('~/.ansibullbot/cache/module_indexer.db')
```
pilou-: `--cachedir` value should be used for parent database directory. `ansible_module_indexer.db` could be used  instead of `module_indexer.db`.

ansibullbot/utils/moduletools.py
```diff
@@ -80,6 +118,16 @@ def manage_checkout(self):
             changed = self.update_checkout()
         return changed
 
+    def get_files(self):
+
+        cmd = 'find %s' % self.checkoutdir
```
pilou-: `self.files` should not contains directories (nor `.pyc`). Should not `git ls-files` used instead ?

ansibullbot/utils/moduletools.py
```diff
@@ -418,13 +469,102 @@ def last_commit_for_file(self, filepath):
         return so.strip()
 
     def get_module_blames(self):
+
+        logging.debug('build email cache')
+        emails_cache = self.session.query(Email)
+        emails_cache = [(x.email,x.login) for x in emails_cache]
+        self.emails_cache = dict(emails_cache)
```
pilou-: `self.emailmap` must be removed (because being replaced by `self.emails_cache`).

In `extract_github_id` method, `self.emails_cache` must be used (instead of `self.emailmap`).

PR #777, opened 2017-09-19
Fix test_template_extractor test_2

samdoran:
Create function for scrubbing bad characters with a few exceptions based on regular expressions.
Add additional test for when Markdown headers have the option `####` at the end of the line.

### DIFF
--- a/ansibullbot/utils/extractors.py
+++ b/ansibullbot/utils/extractors.py
@@ -193,7 +193,6 @@ def extract_template_data(body, issue_number=None, issue_class='issue', sections
         # remove pre-ceding and trailing newlines (AGAIN)
         v = v.strip()
 
-        # import epdb; epdb.st()
         # clean more on critical sections
         if 'step' not in k and 'result' not in k:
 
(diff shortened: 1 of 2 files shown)


ansibullbot/utils/extractors.py
```diff
@@ -193,8 +193,9 @@ def extract_template_data(body, issue_number=None, issue_class='issue', sections
         # remove pre-ceding and trailing newlines (AGAIN)
         v = v.strip()
 
+        # import epdb; epdb.st()
         # clean more on critical sections
-        if k != 'summary' and 'step' not in k and 'result' not in k:
```
samdoran: Not skipping the `summary` section has several benefits:
- extra Markdown header syntax at the end of the line is properly scrubbed
- section headers that end in anything, really, get nicely cleaned up

I'm not sure why this was skipped before. Is there harm in allowing `summary` to be cleaned up?

jctanner: Only to reduce cpu costs.

samdoran: Hmmmm... Well I need some of the scrubbing that happens here for the summary field. Should I write another section just for processing the summary that does the bare minimum?

PR #781, opened 2017-09-29
Add auto-labeling to the list of things ansibot does

dagwieers:
Also, I think we should rename "ansibullbot" into "ansibot" in the documentation, as that's what most people will see. Using a different name for the same thing will make people wonder whether this is something different.


pilou-:
About `s/ansibullbot/ansibot/`, opposite was recently done https://github.com/ansible/ansible/pull/28368/commits/e82be4a4143cfcf9e0438f7bcb1d5b53f9a1c114.

PR #790, opened 2017-10-17
Add parameter to use an alternative botmeta file

jctanner:


jctanner:
@samdoran @pilou- this should help with future testing.

PR #804, opened 2017-10-25
Attributes initialization v2

pilou-:
- Add Pylint plugins for triagers (due to dynamic attributes)
- move common triagers code to `defaulttriager.py`
- remove useless `self.args`

The next step will be:
- remove code which handles old repositories (`ansible-modules-core` and `ansible-modules-extras`)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,2 @@
+[MASTER]
+load-plugins=pylint_plugins/triagers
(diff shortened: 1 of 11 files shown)


jctanner:
We can't completely remove the code specific to the module repos, because people are still opening issues/prs there. If you meant that you would be moving the code, that's okay.

pilou-:
This pull-request doesn't remove the code specific to the module repos. The main change is about removing `self.args`.

PR #826, opened 2017-11-30
Deprecated modules: fix handling of maintainers

pilou-:
`̀BOTMETA.yml` entries applying to deprecated modules were ignored.

For example [`modules/cloud/azure/azure.py`](https://github.com/ansible/ansible/blob/35acae7ea7b62aec3d33c6b96630f750e5b2b43b/.github/BOTMETA.yml#L156) entry was ignored because the module has been deprecated and has been renamed to [`modules/cloud/azure/_azure.py`](https://github.com/ansible/ansible/tree/35acae7ea7b62aec3d33c6b96630f750e5b2b43b/lib/ansible/modules/cloud/azure).

This pull-request allows to take in account entries related to deprecated modules.

Not sure if there is any advantage to handle deprecated modules like [that](https://github.com/ansible/ansibullbot/blob/76cd2da9ab0d50221c1d69795661db28fe9c9f8e/ansibullbot/utils/moduletools.py#L393): I mean instead of having two entries in `self.modules` for each deprecated module, only the one corresponding to the deprecated path could be kept (then `BOTMETA.yml` should be updated when a module become deprecated).

Unit tests provided.

### DIFF
--- a/ansibullbot/triagers/plugins/notifications.py
+++ b/ansibullbot/triagers/plugins/notifications.py
@@ -12,6 +12,9 @@ def get_notification_facts(issuewrapper, meta, file_indexer):
         'to_assign': []
     }
 
+    if iw.is_pullrequest() and iw.merge_commits:
+        return nfacts
+
     # who is assigned?
     current_assignees = iw.assignees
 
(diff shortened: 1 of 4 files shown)


gundalow:
Hey @pilou- What's the plan for this?

pilou-:
@gundalow I guess my plan was to wait for a review ;)

PR #833, opened 2017-12-07
Ensure the shipit commands are on a line by themselves without other …

jctanner:
…text so that accidental merges do not occur


ryansb:
Looks great - will no longer match a 👍 from `@@ -100,12 +100,20 @@`

Issue #837, opened 2017-12-20
Don't set labels or mention people on broken merges


gundalow:
https://github.com/ansible/ansible/pull/34028 shows
* a failed rebase
* Lots of labels added
* lots of people notified

We should detect these type of cases and the bot shouldn't updates labels or `@mentions`

pilou-:
Bot could:
- disable notifications when there are more than 10 commits (PR with more than 10 commits aren't mergeable by the bot, see #835)
- add `needs_rebase` label to these pull-request (because submitter must rebase the PR)

dagwieers:
This is a duplicate of https://github.com/ansible/ansibullbot/issues/787

MarkusTeufelberger:
Another one: https://github.com/ansible/ansible/pull/34264

Issue #787, opened 2017-10-12
Prevent notification storms on corrupted/improper PRs


dagwieers:
So almost daily we get some broken PRs with lots of existing commits in a new PR. Often the owner of the branch did something wrong and has no clue to resolve it and eventually just closes the ticket. But the real damage is already done at the time, as everybody involved with any of these commits (directly or indirectly) are getting mailed by ansibot.

Examples:
- ansible/ansible#31505
- ansible/ansible#31654
- ansible/ansible#31665

Either we should be able to identify these corrupt PRs, or at least prevent to send out notifications to more than an arbitrary number of people e.g. 25 ? Maybe check existing cases for a good treshold ?

pilou-:
Does `more than 30 commits and more than 30 files modified` a good filter for these PR ?

cben:
Proposal: a simple "more than N people at once" limit for automatic notification.

If more than that, ansibot instead of mentioning all the people would explain that it suspects it's a mistake, and will list the usernames it would have cc'd  (without `@`, or within backticks literal), so humans can (A) understand what's happening (B) manually cc all or some people if really intended (eg. large cross-cutting refactoring).

jctanner:
first implementation could be to disable all notifications if a merge commit is detected (something we already check for).

dagwieers:
Another common case recently we also have to consider:
https://github.com/ansible/ansible/issues/33448#issuecomment-348435111

dagwieers:
We are now closing one specific case, but I'd like to also consider a more general solution for other causes (like the one in my previous comment). I want to prevent any more of these cases because they make people want to drop out of any notifications.

gundalow:
Seems to be back. opened https://github.com/ansible/ansibullbot/issues/837

dagwieers:
https://github.com/ansible/ansible/pull/34011

PR #841, opened 2018-01-04
Enhance bad_pr detection slightly and bypass most action creation if true

jctanner:
Fixes https://github.com/ansible/ansibullbot/issues/837
Refixes https://github.com/ansible/ansibullbot/issues/787


gundalow:
is_bad_pr when > 50 files OR > 50 commits sounds like it should catch the issues I've seen.

Thanks for this

PR #855, opened 2018-01-31
Fix needs_info auto-{warn,close}

mkrizek:
https://github.com/ansible/ansible/issues/32432


mkrizek:
Tested on https://github.com/ansible/ansible/issues/32432

PR #859, opened 2018-02-02
more report scripts

jctanner:


gundalow:
What's the plan for this?

Hi when you get a chance, can you please rebase so we get coverage data, see https://github.com/ansible/ansibullbot/pull/1093 for how this would look.
Thanks in advance!

PR #871, opened 2018-02-23
Make bot add performance label

mkrizek:
Allows for easier tracking of performance issues/PRs.

Tested on:
https://github.com/ansible/ansible/issues/36431 (`performance` in `body`)
https://github.com/ansible/ansible/issues/32678 (`performance` in `title`)
https://github.com/ansible/ansible/pull/36626 (no `performance`)

### DIFF
--- a/ansibullbot/triagers/ansible.py
+++ b/ansibullbot/triagers/ansible.py
@@ -72,6 +72,7 @@ from ansibullbot.triagers.plugins.needs_revision import get_needs_revision_facts
 from ansibullbot.triagers.plugins.needs_revision import get_shippable_run_facts
 from ansibullbot.triagers.plugins.contributors import get_contributor_facts
 from ansibullbot.triagers.plugins.notifications import get_notification_facts
+from ansibullbot.triagers.plugins.performance import get_performance_facts
 from ansibullbot.triagers.plugins.py3 import get_python3_facts
 from ansibullbot.triagers.plugins.shipit import automergeable
 from ansibullbot.triagers.plugins.shipit import get_review_facts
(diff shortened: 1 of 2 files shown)


ansibullbot/triagers/plugins/performance.py
```diff
@@ -0,0 +1,19 @@
...
+
+    pfacts = {
+        'is_performance': False
+    }
+
+    body = iw.body.lower()
+    title = iw.title.lower()
+
+    # TODO search in comments too?
+    for data in (body, title):
```
webknjaz: Shorter:
```python
pfacts['is_performance'] = any('performance' in d for d in (body, title))
```

Can also be combined with `pfacts` declaration on line 6

ansibullbot/triagers/ansible.py
```diff
@@ -1113,6 +1114,11 @@ def create_actions(self, iw, actions):
             if comment not in actions.comments:
                 actions.comments.append(comment)
 
+        # performance issue/PR
+        if self.meta['is_performance']:
+            if 'performance' not in iw.labels:
```
webknjaz: Why not use one `if` conditional instead of double indenting body with two? Clause would perfectly fit in a single line.

ansibullbot/triagers/plugins/performance.py
webknjaz: Also, `any` adds a bit of semantics to the code, which improves readability. I would likely to understand it much faster when I don't have to scan through 4 lines of loop.

ansibullbot/triagers/ansible.py
mkrizek: Mostly to keep style consistent with the rest of the code.

ansibullbot/triagers/plugins/performance.py
mkrizek: I was going to say that it decreases readability after reading the first comment :) Matter of taste I guess, I'll let @jctanner decide.

PR #872, opened 2018-02-26
`documentation report` & `docs pull request` should add label:docs

gundalow:
Just use a single label `docs` not `docs_pull_request` or `docs_request`
See https://github.com/ansible/ansible/pull/36757

PR #875, opened 2018-02-27
Add script for replacing labels

mkrizek:
Adding a weird script as an addition to https://github.com/ansible/ansibullbot/pull/872, for a review before running it. Could be re-worked for similar purpose in future.

Tested on:
https://github.com/ansible/ansible/issues/12445 (apparently I suck at `requests` hence the several tries to do it right)

### DIFF
--- a/scripts/list_bug_report.sh
+++ b/scripts/list_bug_report.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+URL='https://github.com/ansible/ansible/issues?q=label%3Abug_report'
+PYTHONPATH=$(pwd) scripts/scrape_github_issues_url $URL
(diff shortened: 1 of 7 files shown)


scripts/replace_docs_labels.py
```diff
@@ -0,0 +1,55 @@
...
+
+            for label in REMOVE_LABELS:
+                if label in labels:
+                    removed_labels.add(label)
+
+            for label in removed_labels:
+                url = LABEL_URL_FMT.format(number, '/'+label)
+                r = requests.delete(url, headers=HEADERS)
+                if r.status_code != 200:
+                    import epdb; epdb.st()
```
webknjaz: Of hide this behind debug flag and add some logging.

PR #910, opened 2018-05-15
Link to devel docs

gundalow:
As PRs go into the devel branch, contributors should be following the
devel guides, rather than the "latest" stable release.

Ensure ensure that we link to direct pages, rather than requiring
redirects.


jctanner:
Thanks @gundalow !

Issue #393, opened 2017-03-01
small_patch label


jctanner:
Determine criteria for a "small" patch, and autolabel it.

jctanner:
@abadger please detail your thoughts on what "small" means.

abadger:
Starter heuristic might be: 
* no more than two hunks of code are changed.
* Each hunk should be at most three lines changed.

It's not a perfect heuristic for deciding if something is an easyfix but it does let us filter for things that might be easy to review.

PR #925, opened 2018-05-22
[RFC] Add small_patch label

mkrizek:
Fixes https://github.com/ansible/ansibullbot/issues/393

Will add tests once we agree on the starter heuristic.

@abadger proposed the following in the issue:
> * no more than two hunks of code are changed.
> * Each hunk should be at most three lines changed.

I could not figure out how to count changes in each chunk in a sane way so the code does the following:
* any new or deleted files in a PR are not classified as `small_patch`
* each modified file in a PR can change 6 lines at most (additions and deletions)
* tests are not counted as a change, see PR I tested on below (but PRs only containing tests are not `small_patch`)
* PR can have two chunks of changed code at most

So if I have not confused myself, the biggest `small_patch` could be two changed files with 6 added lines in each. Thinking about it, 6 sounds more that it should be though.

Tested on:
* https://github.com/ansible/ansible/pull/40338
* https://github.com/ansible/ansible/pull/36870

Thoughts?


jctanner:
@mkrizek need to rebase this

mkrizek:
Rebased.

Issue #934, opened 2018-05-24
bot is attemping to use 'zuul.openstack.org' as a shippable jobid


jctanner:
```
################################ TRACEBACKS #########################
main()
File "/home/ansibot/ansibullbot/triage_ansible.py", line 38, in main
AnsibleTriage().start()
File "/home/ansibot/ansibullbot/ansibullbot/triagers/defaulttriager.py", line 180, in start
self.loop()
File "/home/ansibot/ansibullbot/ansibullbot/triagers/defaulttriager.py", line 290, in loop
self.run()
File "/home/ansibot/ansibullbot/ansibullbot/triagers/ansible.py", line 505, in run
action_meta = self.apply_actions(iw, actions)
File "/home/ansibot/ansibullbot/ansibullbot/triagers/ansible.py", line 1417, in apply_actions
return super(AnsibleTriage, self).apply_actions(iw, actions)
File "/home/ansibot/ansibullbot/ansibullbot/triagers/defaulttriager.py", line 322, in apply_actions
self.execute_actions(iw, actions)
File "/home/ansibot/ansibullbot/ansibullbot/triagers/ansible.py", line 2262, in execute_actions
self.SR.cancel(runid)
File "/home/ansibot/ansibullbot/ansibullbot/utils/shippable_api.py", line 360, in cancel
run_id = self.get_run_id(run_number)
File "/home/ansibot/ansibullbot/ansibullbot/utils/shippable_api.py", line 336, in get_run_id
raise Exception("Unable to fetch %r" % run_url)
Exception: Unable to fetch u'https://api.shippable.com/runs?projectIds=573f79d02a8192902e20e34b&isPullRequest=True&runNumbers=zuul.openstack.org'
```

PR #936, opened 2018-05-24
Ignore other CI's other than shipabull's

mkrizek:
Fixes https://github.com/ansible/ansibullbot/issues/934

### DIFF
--- a/ansibullbot/triagers/plugins/ci_rebuild.py
+++ b/ansibullbot/triagers/plugins/ci_rebuild.py
@@ -131,8 +131,12 @@ def get_rebuild_merge_facts(iw, meta, core_team):
     if lc and lc > last_command:
         return rbmerge_meta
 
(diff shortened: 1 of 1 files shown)


ansibullbot/triagers/plugins/ci_rebuild.py
```diff
@@ -7,6 +7,10 @@
 def status_to_date_and_runid(status, keepstate=False):
     """convert pr status to a tuple of date and runid"""
 
+    # https://github.com/ansible/ansibullbot/issues/934
+    if not status.get('creator', {}).get('login', '') == 'shipabull':
```
mattclay: This doesn't look right. Why do we care about the user?

mattclay: Isn't the intent to check to which CI system the status is from (shippable.com), not the user?

Issue #912, opened 2018-05-16
clean up code per reviews


jctanner:
https://github.com/ansible/ansibullbot/pull/906#pullrequestreview-120644028

PR #938, opened 2018-05-25
Code clean up as per review

Akasurde:
Fixes: #912

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/ansibullbot/triagers/plugins/shipit.py
+++ b/ansibullbot/triagers/plugins/shipit.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import itertools
 import logging
 from fnmatch import fnmatch
 from ansibullbot.utils.moduletools import ModuleIndexer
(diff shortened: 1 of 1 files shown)


ansibullbot/triagers/plugins/shipit.py
```diff
@@ -422,25 +423,21 @@ def get_submitter_facts(issuewrapper, meta, module_indexer, component_matcher):
...
+    )
+
+    emails = set(k for k, v in all_emails if v == login)
 
     email_map = \
         component_matcher.gitrepo.get_commits_by_email(emails)
 
-    for k,v in email_map.items():
-        sfacts['submitter_previous_commits'] += v['commit_count']
+    for value in list(email_map.values()):
```
mkrizek: Just curious as to why casting to list explicitly? I know it's because of some changes in Python 3 but does it not work without it?

mkrizek: Look good otherwise :+1:

Akasurde: Yes. It would work without casting. This syntax will work for Python 2 and 3.

PR #954, opened 2018-06-08
Update ansibullbot role

samdoran:


jctanner:
I think I did this already.

PR #960, opened 2018-06-12
[WIP] Handle matching partial file paths better

mkrizek:
https://github.com/ansible/ansible/issues/41384


gundalow:
After discussion with @mkrizek we are closing this as other work has gone on around matching, and this was only a partial fix (and broke other things)

PR #975, opened 2018-07-02
Add docsite_pr and namespace_labels

dagwieers:
At some point it could be useful to add some additional logic, e.g. based on label colors or label descriptions. Or use a blacklist, rather than a whitelist (to protect labels that can cause privilege escalation).

### DIFF
--- a/ansibullbot/triagers/plugins/label_commands.py
+++ b/ansibullbot/triagers/plugins/label_commands.py
@@ -7,24 +7,43 @@ def get_label_command_facts(issuewrapper, meta, module_indexer, core_team=[], va
     add_labels = []
     del_labels = []
 
-    wl = [
-        'needs_triage',
-        'test',
-        'module',
-        'cloud',
+    namespace_labels = [
+        'aci',
+        'avi',
         'aws',
         'azure',
+        'cloud',
+        'cloudstack',
         'digital_ocean',
         'docker',
+        'f5',
         'gce',
-        'openstack',
-        'vmware',
+        'infoblox',
+        'jboss',
+        'meraki',
+        'netapp',
         'networking',
+        'nxos',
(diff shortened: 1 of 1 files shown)


ansibullbot/triagers/plugins/label_commands.py
```diff
@@ -7,24 +7,43 @@ def get_label_command_facts(issuewrapper, meta, module_indexer, core_team=[], va
...
+        'infoblox',
+        'jboss',
+        'meraki',
+        'netapp',
+        'networking',
+        'nxos',
         'openstack',
+        'ovirt',
+        'ucs',
+        'unarchive',
```
jctanner: why is unarchive a namespace label?

dagwieers: It's a "module community" label ? Much like the `xml` label.
It allows to track issues/PRs related to specific modules.

dagwieers: I don't mind removing them from the list if that is an issue.

IMO a blacklist protecting special labels is more convenient than listing namespace labels (and others).

dagwieers: Idea: maybe we could introduce m:<module> for module namespaces and only have ansibot add those when they were pre-created ? This way it's still on-demand, but integrates automatically with ansibot.

jctanner: works for me, shipit.

dagwieers:
@jctanner I added support for module labels (`m:<module>`). So from now on people can add/remove those labels if they were pre-created. I also added `m:unachive` and `m:xml`.

dagwieers:
@jctanner If you want to shipit, merge it ;-)

PR #977, opened 2018-07-06
Ensure owner_pr label remains even when PR needs_revision

dagwieers:
Because the *owner_pr* label was set after the PR was evaluated to be *needs_revision*, it would not be set as soon as CI failed and the PR was tagged as *needs_revision*.

We should ensure that this function does not modify **nmeta** after calls to `return nmeta` otherwise labels would be revoked because of other logic.

PS There's one more such bugfix I will propose in a subsequent PR.

### DIFF
--- a/ansibullbot/triagers/plugins/shipit.py
+++ b/ansibullbot/triagers/plugins/shipit.py
@@ -242,6 +242,13 @@ def get_shipit_facts(issuewrapper, meta, module_indexer, core_team=[], botnames=
             nmeta['owner_pr'] = True
             return nmeta
 
+    modules_files_owned = 0
+    if not meta['is_new_module']:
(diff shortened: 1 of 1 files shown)


ansibullbot/triagers/plugins/shipit.py
```diff
@@ -242,6 +242,13 @@ def get_shipit_facts(issuewrapper, meta, module_indexer, core_team=[], botnames=
             nmeta['owner_pr'] = True
             return nmeta
```
dagwieers: This call is wrong here as well and could cause subsequent nmeta-changes to get lost, or make shipits fail to work.

dagwieers: Fixed in #978

PR #978, opened 2018-07-06
Because `return nmeta` shipits would fail

dagwieers:
Exiting the function prematurely would trump shipits evaluation for module_utils PR with *owner_pr* label.


dagwieers:
It seems that this triggers another issue which I don't understand. Theoretically the code path should be identical whether or not the module_utils PR is an *owner_pr* or not. Now the function returns when it is an *owner_pr* and *module_utils*. It feels wrong.

mkrizek:
Superseded by #1046.

PR #990, opened 2018-07-31
Autountriage if created by Core

### LINKED ISSUE #989: We should not have to triage issues/PRs created by Core members

webknjaz:
Fixes #989

### DIFF
--- a/ansibullbot/triagers/ansible.py
+++ b/ansibullbot/triagers/ansible.py
@@ -2179,8 +2179,9 @@ class AnsibleTriage(DefaultTriager):
         return {'waiting_on': wo}
 
     def get_triage_facts(self, issuewrapper, meta):
+        iw = issuewrapper
         tfacts = {
-            'maintainer_triaged': False
+            'maintainer_triaged': iw.issue.user.login in self.ansible_core_team,
         }
 
         '''
@@ -2197,22 +2198,22 @@ class AnsibleTriage(DefaultTriager):
         if not meta.get('component_maintainers'):
             return tfacts
 
-        iw = issuewrapper
         #maintainers = [x for x in meta['module_match']['maintainers']]
         maintainers = meta['component_maintainers'][:]
         maintainers += [x for x in self.ansible_core_team]
         maintainers = [x for x in maintainers if x != iw.submitter]
         maintainers = sorted(set(maintainers))
-        if iw.history.has_commented(maintainers):
-            tfacts['maintainer_triaged'] = True
-        elif iw.history.has_labeled(maintainers):
-            tfacts['maintainer_triaged'] = True
-        elif iw.history.has_unlabeled(maintainers):
-            tfacts['maintainer_triaged'] = True
-        elif iw.is_pullrequest() and iw.history.has_reviewed(maintainers):
-            tfacts['maintainer_triaged'] = True
-        elif iw.history.was_self_assigned():
-            tfacts['maintainer_triaged'] = True
+
+        tfacts['maintainer_triaged'] |= (
+            iw.history.has_commented(maintainers) or
+            iw.history.has_labeled(maintainers) or
+            iw.history.has_unlabeled(maintainers) or
+            (
+                iw.is_pullrequest() and
+                iw.history.has_reviewed(maintainers)
+            ) or
+            iw.history.was_self_assigned()
+        )
 
         return tfacts
 
--- a/ansibullbot/triagers/ansible.py
+++ b/ansibullbot/triagers/ansible.py
@@ -2181,7 +2181,7 @@ class AnsibleTriage(DefaultTriager):
     def get_triage_facts(self, issuewrapper, meta):
         iw = issuewrapper
         tfacts = {
-            'maintainer_triaged': iw.issue.user.login in self.ansible_core_team,
+            'maintainer_triaged': iw.submitter in self.ansible_core_team,
         }
 
         '''


ansibullbot/triagers/ansible.py
```diff
@@ -2179,8 +2179,9 @@ def waiting_on(self, issuewrapper, meta):
         return {'waiting_on': wo}
 
     def get_triage_facts(self, issuewrapper, meta):
+        iw = issuewrapper
         tfacts = {
-            'maintainer_triaged': False
+            'maintainer_triaged': iw.issue.user.login in self.ansible_core_team,
```
mkrizek: I don't think `iw.issue.user.login` exists. You probably want `iw.submitter` instead?

webknjaz: Dunno, I've followed doc:
* https://pygithub.readthedocs.io/en/latest/github_objects/Issue.html (`iw.issue`)
* https://pygithub.readthedocs.io/en/latest/github_objects/Issue.html#github.Issue.Issue.user (`iw.issue.user`) -> https://pygithub.readthedocs.io/en/latest/github_objects/NamedUser.html
* https://pygithub.readthedocs.io/en/latest/github_objects/NamedUser.html#github.NamedUser.NamedUser.login (`iw.issue.user.login`)

And there's nothing called `submitter` in that doc.

webknjaz: Oh, I see. Yes, `iw.submitter` from https://github.com/ansible/ansibullbot/blob/master/ansibullbot/wrappers/defaultwrapper.py#L853 should work.

webknjaz:
@mkrizek I believe, this is mergeable.

ryansb:
+0 for this change from me. I do look for core PRs first when doing triage, so I would likely see them later/possibly not at all if they didn't have that label. If this is merged, I can make a query instead that would give me the same result. 

> I'm one of those people that uses triage to note core PRs to review. Unless you want to have to ping a core member to review every time, this may not be good.
> 
> - @bcoca

ansibullbot/triagers/ansible.py
```diff
@@ -2179,8 +2179,9 @@ def waiting_on(self, issuewrapper, meta):
         return {'waiting_on': wo}
 
     def get_triage_facts(self, issuewrapper, meta):
+        iw = issuewrapper
         tfacts = {
-            'maintainer_triaged': False
+            'maintainer_triaged': iw.submitter in self.ansible_core_team,
```
webknjaz: *Remark:* might need to have ` and not `@jimi-c

abadger:
I'd be for this.... Our goal should be to get human's out of the triage process.  It really should be the core team member's responsibility to push their PRs to the relevant people when they're ready rather than relying on whoever is triaging that day to be a go between for the communication.

bcoca:
if i'm only one, ignore the objection, i'll adapt

sivel:
Apologies for not commenting here as well.  I am also -1 to this change.  It seems bcoca and I share a similar workflow here.

webknjaz:
@sivel what about workarounds? separate label for that workflow. it seems to me that you both abuse this label despite its original purpose

samdoran:
I'm -1 to this change as I like to see core team member issue/PRs in the triage process. It's nice to sanity check as well as actually review or merge them. Plus I know someone else will look at my submissions to ensure I submitted everything correctly.

I (maybe incorrectly) assume that when I submit PRs they will get seen during triage and the appropriate person will review them. Otherwise I feel like I have to nag someone to get it reviewed, which makes me feel bothersome to others.

abadger:
@samdoran your assumption is incorrect.  You do have to nag the right person to get it reviewed.

samdoran:
@abadger Thanks for the clarification.

webknjaz:
The submitter must decide themselves whom they want to ask to do a review. Otherwise, it's noise. When doing triage I almost never notify anyone, because I have no way of knowing what submitter wanted. So it's just a garbage in my workflow.
If you insist on keeping this label I'll need some other label for myself to have a filter for skipping frustrating stuff.

sivel:
I don't always need a specific person to review.  Additionally, who is to say that what I submitted in the issue/pr is enough information for a person to later look at or work the issue.

When I triage issues/prs I ensure that the issue/pr is properly formatted has all of the pertinent data needed by another person to work the issue/pr.

You would expect a core team member to give that data, but I am biased, and may not understand the needs of another person in handling the issue later.

webknjaz:
@abadger @sivel maybe a bit unrelated, but what do you think of using [`.github/CODEOWNERS`](https://blog.github.com/2017-07-06-introducing-code-owners/) to autoinvite relevant ppl for review? I've tried using it in other projects and it makes ppl asking for review easier: ones who are experts on certain files in repo are asked to review by GH, reducing the load on triager, who needs to somehow try to identify a person responsible for that area every time.

mkrizek:
> I'm -1 to this change as I like to see core team member issue/PRs in the triage process. It's nice to sanity check as well as actually review or merge them. Plus I know someone else will look at my submissions to ensure I submitted everything correctly.

@samdoran agreed.

FWIW, at this moment, there are 72 issues with `needs_triage`, if my math is correct, 7 of those are from core ( + network) members. I don't see this as a huge additional load on a person doing triage.

webknjaz:
> I don't see this as a huge additional load on a person doing triage.

It requires a different approach to processing those 10%. It implies a different level of trust to such contribution. Just because of this it causes context switching, which generates bits of frustration.
That's why I need to somehow separate them.

bcoca:
@webknjaz that probably explains it, i never look at the user when doing triage or even when reviewing the PR. I think it makes me more objective if i take the 'person' out of the equation .. at least until i need to communicate with them.

gundalow:
So based on above and discussion in #1024 is the answer
Don't label `needs_triage` if:
* created by person with commit 
AND
* it's a backport

webknjaz:
@gundalow I think "created with commit" is irrelevant to backport. Since it's being reviewed by RMs and need to be excluded from triage I guess.

abadger:
Yeah.  The bot could do other (does it already?) Checks but those would be unrelated to committer... Is the backport's initial commit merged?  (If not is there an open PR for it?) Is there a changelog fragment?  Does the backport differ from the devel commit?

mkrizek:
It was decided by the core team that this is not something we want implemented at this point.

PR #999, opened 2018-08-16
fix compat with POSIX shell

pilou-:
Fix compatibility with shell with POSIX features only (eg `dash`):
* use `.` instead of `source`
  Fix this error:

      /bin/sh: 1: source: not found

* and set `cwd` (otherwise [`HACKING_DIR`](https://github.com/ansible/ansible/blob/ac895828c1a57e8054eb80969d214fad399d252e/hacking/env-setup#L42) default to `PWD` when POSIX shell is used and then `ANSIBLE_HOME` is wrong)


mkrizek:
Closing in favor of https://github.com/ansible/ansibullbot/pull/1002.

PR #1001, opened 2018-08-17
component matcher: remove unused code

pilou-:
Remove unused code

PR #1000, opened 2018-08-16
Install Ansible dependencies too

pilou-:
`ansible-doc` is used by `AnsibleComponentMatcher`:
https://github.com/ansible/ansibullbot/blob/149f8e099a8d5822805418c62c73f3fd6256371f/ansibullbot/utils/component_tools.py#L162
then Ansible dependencies must be installed too.

It should fix shippable errors reported by shippable build of #999:
```
ERROR: test_reduce_filepaths (tests.unit.utils.test_component_tools.TestComponentMatcher)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/src/github.com/ansible/ansibullbot/tests/unit/utils/test_component_tools.py", line 116, in test_reduce_filepaths
    CM = get_component_matcher()
  File "/root/src/github.com/ansible/ansibullbot/tests/unit/utils/test_component_tools.py", line 106, in get_component_matcher
    module_indexer=MI
  File "/root/src/github.com/ansible/ansibullbot/ansibullbot/utils/component_tools.py", line 112, in __init__
    self.update()
  File "/root/src/github.com/ansible/ansibullbot/ansibullbot/utils/component_tools.py", line 118, in update
    self.index_files()
  File "/root/src/github.com/ansible/ansibullbot/ansibullbot/utils/component_tools.py", line 166, in index_files
    raise Exception("'ansible-doc' command failed (%s, %s %s)" % (rc, so, se))
Exception: 'ansible-doc' command failed (250, running egg_info
creating lib/ansible.egg-info
writing requirements to lib/ansible.egg-info/requires.txt
writing lib/ansible.egg-info/PKG-INFO
writing top-level names to lib/ansible.egg-info/top_level.txt
writing dependency_links to lib/ansible.egg-info/dependency_links.txt
writing manifest file 'lib/ansible.egg-info/SOURCES.txt'
reading manifest file 'lib/ansible.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'lib/ansible.egg-info/SOURCES.txt'

Setting up Ansible to run out of checkout...

PATH=/tmp/tmpy_ZnqF/ansible/bin:/tmp/tmpy_ZnqF/ansible/test/runner:/root/venv/2.7/bin:/root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/root/usr/local/bin
PYTHONPATH=/tmp/tmpy_ZnqF/ansible/lib:
[...]

PR #1002, opened 2018-08-17
Update component matcher tests

pilou-:
Depends on #999 (not merged yet), #1000 (not merged yet) and #1001 (not merged yet): the eight first commits come from these 3 pull requests.

New commits are the last five:
* e52210a: check return code of `ansible-doc`
* db0db84: use 'devel' tarball
* 7b7fb2b: update renamed paths
* 41ff849 : disable non working test
* 43001da: use a shallow clone instead of a tarball


mkrizek:
@pilou- should we just squash and merge this one in favor of #999, #1000 and #1001?

tests/unit/utils/test_component_tools.py
```diff
@@ -129,10 +55,12 @@ def test_search_by_filepath(self):
                 'lib/ansible/modules/network/ios/ios_facts.py',
                 'lib/ansible/modules/network/ios/ios_facts.py'
             ],
-            'json_query': [
-                'lib/ansible/plugins/filter/json_query.py',
-                'lib/ansible/plugins/filter/json_query.py'
-            ],
+            # Doesn't work, lib/ansible/modules/network/nso/nso_query.py is
```
mkrizek: So has this never worked?

pilou-: It works with [2017-10-24 Ansible source tree](https://github.com/ansible/ansibullbot/blob/fae5f7c4af5d170ef101fe58a284a61ea7bca8e8/tests/unit/utils/test_component_tools.py#L57): `nso_query` module was added after ([2017-12-22](https://github.com/ansible/ansible/pull/34188)).

I would rather use an up to date tree (meaning `devel` branch) rather than using an out of date tree. Advantage: new failures like this one are detected.

The next step could be to:
* mock either the `ansible-doc` command or the ansible repository content
* add integration test in order to check that ansibullbot is able to call `ansible-doc` (in order to avoid problem like e52210adea8032d996ff333322b6a4922e01e6ee)

mkrizek: @pilou- can you please create an issue for the next step?

mkrizek:
Thanks @pilou-!

Issue #971, opened 2018-06-28
Auto-close PRs not using forks


mattclay:
When a PR is opened using a branch in the ansible/ansible repository instead of a fork, the bot should:

- Comment on the PR with a message such as "Please create a new PR using a branch in your fork."
- Close the PR.
- Cancel the CI run for the PR (if still running).
- Cancel the CI run for the branch (if still running). This is a different CI run than the one for the PR.

This will reduce wasted CI resources due to incorrectly created PRs.

PR #1009, opened 2018-08-29
Auto-close PRs not using forks

mkrizek:
Fixes https://github.com/ansible/ansibullbot/issues/971


mkrizek:
Good catch @gundalow, thanks! That was my `git add` fail :(

PR #1010, opened 2018-08-31
BOTMETA: propagate 'labels', 'support' and 'supported_by' key

pilou-:
Propagate `labels`, `support` and `supported_by` keys:
* `labels`: entry inherits from all ancestors
* `support`/`supported_by`: value isn't cumulated, entry inherits from the nearest ancestor and only if the entry doesn't set this keyword

cc @gundalow 

Unit tests provided.

### DIFF
--- a/ansibullbot/parsers/botmetadata.py
+++ b/ansibullbot/parsers/botmetadata.py
@@ -94,21 +94,33 @@ class BotMetadataParser(object):
                 data['macros'][k] = names
             return data
 
-        def _propagate(files, top, child, field):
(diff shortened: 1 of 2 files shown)


gundalow:
On hold till @pilou- confirms that the work here is OK wrt https://github.com/ansible/ansible/pull/44903#discussion_r214805076

gundalow:
Confirmed,

> For "ANSIBLE_METADATA should override BOTMETA (for support/supported_by)" yes a new ansibullbot PR is required

PR #1012, opened 2018-09-03
Add yum-cron/update-ansibullbot.yml

zeniri:
Added become as True

### DIFF
--- a/playbooks/update-ansibullbot.yml
+++ b/playbooks/update-ansibullbot.yml
@@ -3,5 +3,7 @@
   become: True
 
   roles:
-    - role: ansibullbot
+    - roles: ansibullbot
       ansibullbot_action: update
+    - yum-cron  
+ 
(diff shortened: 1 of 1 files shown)


playbooks/update-ansibullbot.yml
```diff
@@ -1,7 +1,9 @@
 - name: Update Ansibullbot
   hosts: ansibullbot
-  become: yes
+  become: True
 
   roles:
-    - role: ansibullbot
+    - roles: ansibullbot
```
samdoran: This in the incorrect key. It should be `role:`.

samdoran:
The [YAML spec](http://yaml.org/type/bool.html) is pretty flexible with regards to booleans. Using True/False or yes/no is mainly a stylistic preference. I prefer yes/no in YAML since it's a bit more human friendly.

Issue #1025, opened 2018-09-12
iw.pullrequest.updated_at.isoformat() results in 500


mkrizek:
This happens occasionally, probably nothing we can do about it but noting here in case we want to at least investigate.

```
2018-09-11 21:42:58,964 DEBUG GET https://api.github.com/repos/ansible/ansible/pulls/33469 {'Authorization': 'token (oauth token removed)', 'User-Agent': 'PyGithub/Python'} null ==> 500 {'status': '500 Internal 
Server Error', 'x-accepted-oauth-scopes': '', 'content-length': '0', 'x-github-media-type': 'github.v3; format=json', 'content-security-policy': "default-src 'none'", 'x-content-type-options': 'nosniff', 'access
-control-expose-headers': 'ETag, Link, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval', 'x-github-request-id': '9
982:21FF:8E3087:1408B94:5B9836E1', 'strict-transport-security': 'max-age=31536000; includeSubdomains; preload', 'x-oauth-scopes': 'public_repo, read:org, repo:status, repo_deployment', 'x-ratelimit-remaining': '
3510', 'server': 'GitHub.com', 'x-ratelimit-limit': '5000', 'x-runtime-rack': '1.862093', 'date': 'Tue, 11 Sep 2018 21:42:58 GMT', 'x-frame-options': 'deny', 'access-control-allow-origin': '*', 'referrer-policy'
: 'origin-when-cross-origin, strict-origin-when-cross-origin', 'content-type': 'application/json; charset=utf-8', 'x-xss-protection': '1; mode=block', 'x-ratelimit-reset': '1536703692'} 
2018-09-11 21:42:58,966 ERROR 500 None
2018-09-11 21:42:58,967 ERROR Uncaught exception
Traceback (most recent call last):
  File "/home/ansibot/ansibullbot/triage_ansible.py", line 42, in <module>
    main()
  File "/home/ansibot/ansibullbot/triage_ansible.py", line 38, in main
    AnsibleTriage().start()
  File "/home/ansibot/ansibullbot/ansibullbot/triagers/defaulttriager.py", line 180, in start
    self.loop()
  File "/home/ansibot/ansibullbot/ansibullbot/triagers/defaulttriager.py", line 290, in loop
    self.run()
[...]

PR #1028, opened 2018-09-13
Handle 500 ISEs from github's API.

jctanner:
* config to set the github url to allow for testing mocked endpoints
* github_sim flask script to simulate a minimal part of the github rest api
* sleep 2 minutes if 500 ISE is encountered

Fixes https://github.com/ansible/ansibullbot/issues/1025

### DIFF
--- a/ansibullbot/decorators/github.py
+++ b/ansibullbot/decorators/github.py
@@ -19,6 +19,11 @@ import ansibullbot.constants as C
 
 
 def get_rate_limit():
+    url = C.DEFAULT_GITHUB_URL
+    if not url:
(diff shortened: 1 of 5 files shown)


ansibullbot/decorators/github.py
```diff
@@ -171,10 +178,13 @@ def inner(*args, **kwargs):
                             raise Exception('unhandled message type')
                 elif isinstance(e, httplib.IncompleteRead):
                     # https://github.com/ansible/ansibullbot/issues/593
-                    stime = 5
+                    stime = 2*60
```
webknjaz: I traditionally suggest using exponential backoff, instead of linear.

PR #1043, opened 2018-09-20
Add link to maintainer identification description

techtonik:


techtonik:
Actual detail is not sufficient enough to be considered a documentation. There is only a link to source and data file with no description of actual identification logic. I am not the one who can restore that details.

mkrizek:
@techtonik Thank you for the pull request. I agree with @gundalow that this is not a proper way to document this. I am going to go ahead and close this in favor of re-opening the original issue as a "documentation request".

Thanks!

PR #1050, opened 2018-09-24
sanity tests no longer allows `authors:`

dagwieers:
The validate-modules sanity tests no longer allow `authors:` to be defined.
And none of the modules have this specified at this time, so this can go.

### DIFF
--- a/ansibullbot/utils/extractors.py
+++ b/ansibullbot/utils/extractors.py
@@ -501,10 +501,6 @@ class ModuleExtractor(object):
         if not ydata:
             return []
 
-        # sometimes the field is 'author', sometimes it is 'authors'
-        if u'authors' in ydata:
-            ydata[u'author'] = ydata[u'authors']
-
         # quit if the key was not found
         if u'author' not in ydata:
             return []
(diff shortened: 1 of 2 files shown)


ansibullbot/utils/extractors.py
```diff
@@ -498,10 +498,6 @@ def get_module_authors(self):
         if not ydata:
             return []
 
-        # sometimes the field is 'author', sometimes it is 'authors'
-        if 'authors' in ydata:
-            ydata['author'] = ydata['authors']
-
```
pilou-: Should not the duplicated code below be updated too ? https://github.com/ansible/ansibullbot/blob/de3fde5704e113d30babaa5340dcd7f1c6b9ea7f/ansibullbot/utils/moduletools.py#L899-L902

mkrizek: :+1:

webknjaz:
Hey @dagwieers this needs rebasing

dagwieers:
@webknjaz Please merge :-)

gundalow:
Dag, what do you mean by
> The validate-modules sanity tests no longer allow authors: to be defined.

webknjaz:
@gundalow I think he means that those linters present in ansible/ansible were used to allow both `author:` and `authors:` forms of metadata key. But with the recent change that linter allows only `author:` so we don't have to shim a now impossible case on the bot side.

PR #1051, opened 2018-09-25
Make it py3-compatible

webknjaz:
Now it should run under 3.7


shippable.yml
```diff
@@ -1,7 +1,12 @@
 language: python
 python:
   - 2.7
+  - 3.7
 build:
+  pre_ci_boot:
+    image_name: drydock/u16pytall
+    image_tag: v6.8.4
```
webknjaz: @mattclay could you please verify that it's a right way to go?

shippable.yml
```diff
@@ -1,6 +1,7 @@
 language: python
 python:
   - 2.7
+  - 3.7
```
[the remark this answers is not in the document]
mkrizek: Yes, but do we want to throw our shippable resources at it is what I meant.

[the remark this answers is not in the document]
webknjaz: Oh, we can test 3.7 under Travis CI if that's a case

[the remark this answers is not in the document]
mattclay: Looks good. This will consume very little resources compared to ansible/ansible activity.

shippable.yml
```diff
@@ -1,7 +1,12 @@
...
 build:
+  pre_ci_boot:
+    image_name: drydock/u16pytall
+    image_tag: v6.8.4
+    pull: true
```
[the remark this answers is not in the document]
mattclay: This is what we use for ansible/ansible:

```yaml
build:
  pre_ci_boot:
    image_name: drydock/u16pytall
    image_tag: master
    pull: false
    options: "--privileged=false --net=bridge"
```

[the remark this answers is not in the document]
webknjaz: @mattclay I think default doesn't have Python 3.7 in it

ansibullbot/utils/component_tools.py
```diff
@@ -136,8 +138,8 @@ def index_files(self):
             if os.path.isdir(fn):
                 continue
             mname = os.path.basename(fn)
-            mname = mname.replace('.py', '').replace('.ps1', '')
-            if mname.startswith('__'):
+            mname = mname.replace(b'.py', b'').replace(b'.ps1', b'')
+            if mname.startswith(b'__'):
```
abadger: You should establish some convention to show that self.gitrepo.module_files is a list of byte strings.  In ansible proper we name variables that hold byte strings with a b_ prefix.  So: self.gitrepo.b_module_files

(If you don't do this, at some point someone will put text strings in instead of byte strings and then this code will break.)

ansibullbot/utils/component_tools.py
```diff
@@ -154,9 +156,9 @@ def index_files(self):
 
         # make a list of names by enumerating the files
         self.MODULE_NAMES = [os.path.basename(x) for x in self.gitrepo.module_files]
-        self.MODULE_NAMES = [x for x in self.MODULE_NAMES if x.endswith('.py') or x.endswith('.ps1')]
-        self.MODULE_NAMES = [x.replace('.ps1', '').replace('.py', '') for x in self.MODULE_NAMES]
-        self.MODULE_NAMES = [x for x in self.MODULE_NAMES if not x.startswith('__')]
+        self.MODULE_NAMES = [x for x in self.MODULE_NAMES if x.endswith((b'.py', b'.ps1'))]
+        self.MODULE_NAMES = [x.replace(b'.ps1', b'').replace(b'.py', b'') for x in self.MODULE_NAMES]
+        self.MODULE_NAMES = [x for x in self.MODULE_NAMES if not x.startswith(b'__')]
```
abadger: Same comment about a naming convention, here.

ansibullbot/utils/component_tools.py
```diff
@@ -167,21 +169,21 @@ def index_files(self):
...
+            if len(parts) != 2 or checkoutdir not in line.decode():
                 continue
 
             mname = parts[0]
             if mname not in self.MODULE_NAMES:
                 self.MODULE_NAMES.append(mname)
 
             fpath = parts[1]
-            fpath = fpath.replace(checkoutdir + '/', '')
+            fpath = fpath.replace(checkoutdir.encode() + b'/', b'')
```
abadger: Non-ascii characters will break this as well.

abadger:
I only explicitly wrote a few things but you should generalize those to the rest of the code you've written as well.

For the .encode() and .decode() problems.... I would strongly advise you to either make use of an existing to_bytes/to_text function, copy them into your code, or create the minimal version of them.  They make this work much easier.  The version in ansible handles nonstrings which you probably don't need for anything here and also makes use of the surrogateescape error handler if it's available which is something I highly recommend.

ansibullbot/utils/component_tools.py
webknjaz: That's a very good point. I didn't want to remake everything in this PR though.

webknjaz:
@abadger I just wanted to do minimal fixes at first and then think about conventions separately. Regarding encodings, do you think in makes sense to explicitly set `utf-8` as a default encoding at the very beginning?

webknjaz:
I'm remaking things to mostly rely on unicode-strings everywhere now.

ansibullbot/triagers/ansible.py
```diff
@@ -1447,8 +1448,8 @@ def create_actions(self, iw, actions):
             if comment not in actions.comments:
                 actions.comments.append(comment)
 
-        actions.newlabel = sorted(set([x.encode('ascii') for x in actions.newlabel]))
-        actions.unlabel = sorted(set([x.encode('ascii') for x in actions.unlabel]))
+        actions.newlabel = sorted(set([x.encode('ascii').decode('utf-8') for x in actions.newlabel]))
+        actions.unlabel = sorted(set([x.encode('ascii').decode('utf-8') for x in actions.unlabel]))
```
abadger: Not sure if this is the only place where this happens but .doing .encode() and .decode() is usually wrong in general purpose code.  For instance, here, it should probably just be 
``` python
        actions.newlabel = sorted(set(actions.newlabel))
        actions.unlabel = sorted(set(actions.unlabel))
```

The times when it does make sense to both decode and encode:
* converting from one encoding to another.  (In this case, you'd start and end with bytes)
* Making sure that a string only contains characters valid in a specific encoding. (In which case you would encode and decode using the same encoding).

webknjaz: Yeah, I think it was a validation for ascii in labels. In other places I saw `.encode('ascii', 'ignore')` for dropping non-ASCII  characters.

In general, I agree with you that I need to just drop encode/decode from these lines completely.

webknjaz:
@abadger regarding removing encode/decode I think it should remain in places where it's done for the purpose of filtering out non-ascii and it looks it's all places except labels lines where you commented.

As for `to_text`, I'm thinking to add it later.

abadger:
If you think it's there for a reason, two changes should be made

* It should be done when the value enters the system just like you make sure that everything is text at that point.
* You should use the same encoding for both encode and decode rather than utf8

webknjaz:
Good points! I'm currently fixing other places not covered by tests. So will get back to addressing your comments a bit later.

webknjaz:
@jctanner I've disabled two ruamel tests as they are not really friendly to patching. Are they really needed?

This probably can be merged or I can do some further refactoring in this PR. It's just very huge and I don't want to keep fixing conflicts..

ansibullbot/_text_compat.py
```diff
@@ -0,0 +1,278 @@
+# This code is part of Ansible, but is an independent component.
```
pilou-: Since ansible is a [dependency](https://github.com/ansible/ansibullbot/blob/1e459530cc262f088b80140289fbb5dae7087c85/ansibullbot/utils/component_tools.py#L168) of ansibullbot, could not ansibullbot directly import Ansible methods ? Maybe these methods should be moved in a dedicated Python package ? Same comment for `constants.py` file.

webknjaz: It's a non-direct dependency, which resides in a separate env. So no. Also, I'd not want to make these two tightly coupled.

webknjaz:
As for u-strings it's required to run the code under Python 2, because there's no env prepared for running it under Python 3. Also, it's wiser to have a transitional state of the code base as we don't have a complete knowledge of what other compat things I could possibly miss.
And blindly switching versions like this is kinda dangerous given that the test coverage is far from perfect. I've made tests green, but then spent twice as much time tracking down all other failing places and I'm still unsure that it's completely over.

Currently I'm waiting for @abadger's comments to complete this and get it merged ASAP.

jctanner:
please rebase, squash the commits for py3 and separate the changes to the component meta/authors. I will merge afterwards.

Issue #1049, opened 2018-09-24
ansibot does not pick up authors from module


dagwieers:
There seems to be a problem with ansibot not picking up authors from modules.
- https://github.com/ansible/ansible/issues/45970
- https://github.com/ansible/ansible/pull/46060
- https://github.com/ansible/ansible/pull/46068

dagwieers:
cc @willthames @gundalow

dagwieers:
If ansibot returns an empty maintainers-list, we should probably fallback to the default for $team_ansible. And if this is the case, I would propose we remove $team_ansible (and empty container) from BOTMETA.yml ?

PS It is unclear to me what $team_ansible really is used for, and I would suggest using a better marker for this.

jctanner:
This is being caused by component matcher's get_meta_for_file not calling find_module_match and using the metadata provided.

```
(Epdb) pp self.component_matcher.find_module_match('route53')[0].get('authors')
['bpennypacker', u'jimbydamonk']
(Epdb) pp self.component_matcher.get_meta_for_file('lib/ansible/modules/cloud/amazon/route53.py').get('authors')
[]
```

felixfontein:
This still seems to be a problem: ansible/ansible#46583

webknjaz:
@felixfontein that one is double weird:
```json
2018-10-07 13:26:13,475 INFO url: https://github.com/ansible/ansible/issues/46583
2018-10-07 13:26:13,476 INFO title: Ansible v2.70  - Docker Swarm is unable to create the swarm manager 
2018-10-07 13:26:13,476 INFO component[t]: docker_swarm module
2018-10-07 13:26:13,476 INFO component[m]: lib/ansible/modules/cloud/docker/docker_swarm.py
{'assign': [],
 'cancel_ci': False,
 'cancel_ci_branch': False,
 'close': False,
 'close_migrated': False,
 'comments': [],
 'merge': False,
 'newlabel': ['support:core'],
 'open': False,
 'rebuild': False,
 'unassign': [],
 'unlabel': ['needs_maintainer', 'support:community']}
Take recommended actions (y/N/a/R/T/DEBUG)? DEBUG
```

cc @jctanner

webknjaz:
@pilou- I want to merge py3 compat PR first, because it's got a lot in it and I'd like to avoid having to deal with conflicts.

felixfontein:
[...]

PR #1060, opened 2018-10-08
Copy authors and maintainers info from module

webknjaz:
Fixes ansible/ansibullbot#1049

cc @jctanner @gundalow

### DIFF
--- a/tests/unit/utils/test_component_tools.py
+++ b/tests/unit/utils/test_component_tools.py
@@ -86,6 +86,7 @@ class TestComponentMatcher(TestCase):
         self.component_matcher.file_indexer.botmeta = self.component_matcher.BOTMETA = {
             u'files': {
                 u'lib/ansible/modules/packaging/os/yum.py': {
+                    u'ignored': [u'verm666'],
                     u'maintainers': [u'maxamillion'],
                 }
             }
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_component_tools.py
```diff
@@ -103,8 +110,9 @@ def test_get_meta_for_file_powershell(self):
             }
         }
         result = self.component_matcher.get_meta_for_file(u'lib/ansible/modules/windows/win_ping.ps1')
-        self.assertEqual(result[u'labels'], [u'windoez'])
-        self.assertEqual(result[u'maintainers'], [u'jborean93'])
+        assert result[u'labels'] == [u'windoez']
+        expected_maintainers = sorted([u'cchurch', u'jborean93'])
+        assert sorted(result[u'maintainers']) == expected_maintainers
```
gundalow: Is it worth adding a test of `ignore` in here as well?

gundalow:
Tanner said
> Testing ignore seems valid

tests/unit/utils/test_component_tools.py
webknjaz: ignore: done
no authors: impossible
authors + botmeta: dunno how it's different from what we have now

tests/unit/utils/test_component_tools.py
```diff
@@ -86,12 +86,19 @@ def test_get_meta_for_file_pyfile(self):
         self.component_matcher.file_indexer.botmeta = self.component_matcher.BOTMETA = {
             u'files': {
                 u'lib/ansible/modules/packaging/os/yum.py': {
+                    u'ignored': [u'verm666'],
```
webknjaz: @gundalow ^

Issue #972, opened 2018-06-28
status cgi broken if no tracebacks


jctanner:
http://ansibullbot.eng.ansible.com/cgi-bin/ansibot_status.cgi
```
Internal Server Error

The server encountered an internal error or misconfiguration and was unable to complete your request.

Please contact the server administrator at root@localhost to inform them of the time this error occurred, and the actions you performed just before this error.

More information about this error may be available in the server error log.
```

```
[Thu Jun 28 18:44:36.344900 2018] [cgi:error] [pid 9648] [client 63.143.42.245:43092] AH01215: Traceback (most recent call last):, referer: http://ansibullbot.eng.ansible.com/cgi-bin/ansibot_status.cgi
[Thu Jun 28 18:44:36.344943 2018] [cgi:error] [pid 9648] [client 63.143.42.245:43092] AH01215:   File "/var/www/cgi-bin/ansibot_status.cgi", line 156, in <module>, referer: http://ansibullbot.eng.ansible.com/cgi-bin/ansibot_status.cgi
[Thu Jun 28 18:44:36.344951 2018] [cgi:error] [pid 9648] [client 63.143.42.245:43092] AH01215:     rdata += '\\n'.join(tb), referer: http://ansibullbot.eng.ansible.com/cgi-bin/ansibot_status.cgi
[Thu Jun 28 18:44:36.344966 2018] [cgi:error] [pid 9648] [client 63.143.42.245:43092] AH01215: TypeError: sequence item 0: expected string, NoneType found, referer: http://ansibullbot.eng.ansible.com/cgi-bin/ansibot_status.cgi
[Thu Jun 28 18:44:36.359672 2018] [cgi:error] [pid 9648] [client 63.143.42.245:43092] End of script output before headers: ansibot_status.cgi, referer: http://ansibullbot.eng.ansible.com/cgi-bin/ansibot_status.cgi
```

jctanner:
I patched it in place like so:

```
156     rdata += '\n'.join([x for x in tb if x is not None])
```

PR #1063, opened 2018-10-09
Fix internal server error in status.cgi

mkrizek:
Fixes https://github.com/ansible/ansibullbot/issues/972


scripts/ansibot_status.cgi
```diff
@@ -153,7 +153,7 @@ rdata += '<br>'
 rdata += "################################ TRACEBACKS #########################<br>\n"
 for tb in tracebacks:
     rdata += '<pre>'
-    rdata += '\n'.join(tb)
+    rdata += '\n'.join([x for x in tb if x is not None])
```
webknjaz: Nitpick: square brackets ain't actually needed :)

PR #1072, opened 2018-10-12
Codecov.io support

gundalow:
I'm concerned that test coverage isn't that good. Lets see if we can generate some metrics to prove this either way.

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -18,7 +18,7 @@ build:
 
 after_success:
 
-  - codecov --token=272fc7a8-bd55-4e0c-af2f-7d475d0f3c0a
+  - codecov --token=272fc7a8-bd55-4e0c-af2f-7d475d0f3c0a -r :ansible/:ansibullbot
 
   - coverage xml -o $COVERAGE_REPORTS/coverage.xml
 
--- a/shippable.yml
+++ b/shippable.yml
@@ -2,9 +2,6 @@ language: python
 python:
   - 2.7
   - 3.7
-env:
-  - CI_REPORTS=shippable/testresults COVERAGE_REPORTS=shippable/codecoverage
-
 build:
   pre_ci_boot:
     image_name: drydock/u16pytall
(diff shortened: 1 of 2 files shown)


shippable.yml
```diff
@@ -12,3 +15,11 @@ build:
    - shippable_retry pip install -Ur test-requirements.txt
 
    - pytest
+
+after_success:
+
+  - codecov --token=272fc7a8-bd55-4e0c-af2f-7d475d0f3c0a -r ansible/ansibullbot
```
webknjaz: @gundalow it shouldn't require a token for public CIs, just `codecov` command should be enough, it uploads the report where needed.

shippable.yml
```diff
@@ -12,3 +15,11 @@ build:
...
+
+  - coverage xml -o $COVERAGE_REPORTS/coverage.xml
```
webknjaz: What are these supposed to do? pytest already puts a report in that folder. See `pytest.ini`.

test-requirements.txt
```diff
@@ -5,6 +5,7 @@ pip>=9.0
 -r requirements.txt
 
 # Test runner:
+codecov
```
webknjaz: It's not about runner, please put it into a separate section, like `# Reporting:`

webknjaz:
@gundalow currently pytest-cov generates coverage and Shippable even collects it from there, so in each job it's available in a separate tab, like https://app.shippable.com/github/ansible/ansibullbot/runs/731/2/coverage

you'll only need to specify `codecov -f shippable/codecoverage/pytest-cov.xml -X gcov`

Examples:
* https://github.com/cherrypy/cheroot/blob/master/tox.ini#L16 — how I do it in another project
* https://github.com/cherrypy/cheroot/blob/master/.codecov.yml — how you can configure codecov slightly differently from defaults

webknjaz:
@gundalow I've adjusted everything. It's ready to merge.

gundalow:
@webknjaz Thanks!

PR #1078, opened 2018-10-16
Fix unicode json dumping to files

webknjaz:


jctanner:
Please add a unit or integration test to validate this continues to work.

abadger:
This change looks good but the iterencode method itself looks like it may be broken.

* You should use errors="surrogate_or_strict" with to_text so that we don't inject replacement characters.
* You should read the python2 source code for iterencode to make sure the chunks it returns map to whole utf8 characters.  It is quite possible that it is simply chunking bytes without respect to whether it's splitting up bytes that belong to the same utf8 characters when it does that. If the python2 iterencode is naive in this regard, then you'll need to build that logic into this function.

webknjaz:
Fair enough. I'll check that!

webknjaz:
(I have a feeling that those chunks are lines, so I'm not sure whether we'll encounter that "split in the middle of char" issue)

webknjaz:
I don't think I'll split bytes, it yields things wisely: https://github.com/python/cpython/blob/2.7/Lib/json/encoder.py#L412

abadger:
Yeah if it's line based, it should be fine as utf8 newlines will be a single byte

webknjaz:
@abadger it literally doesn't split anything, it just emits lines via generator function.

webknjaz:
CI is green. Ready to merge.

PR #1113, opened 2018-12-04
Add Sentry to Ansibullbot

samdoran:
- add variables for Sentry
- update ansibullbot role to add Sentry support


playbooks/group_vars/ansibullbot.yml
```diff
@@ -21,7 +21,12 @@ ansibullbot_options:
   - skip_no_update_timeout
   - verbose
 
-ansibullbot_receiver_enabled: yes
+
+firewall_allowed_tcp_ports:
+  - 22
+  - 80
```
webknjaz: @samdoran is this just for incoming traffic?

samdoran: Yes.

PR #1117, opened 2018-12-07
Fix TypeError: use bytes (Py3 compat)

pilou-:
Fix this exception which occurs with Python 3.7:

    Traceback (most recent call last):
      File "./triage_ansible.py", line 51, in <module>
        main()
      File "./triage_ansible.py", line 47, in main
        AnsibleTriage().start()
      File "ansibullbot/triagers/defaulttriager.py", line 187, in start
        self.run()
      File "ansibullbot/triagers/ansible.py", line 472, in run
        self.process(iw)
      File "ansibullbot/triagers/ansible.py", line 1910, in process
        shippable=self.SR
      File "ansibullbot/triagers/plugins/needs_revision.py", line 149, in get_needs_revision_facts
        ci_date = get_last_shippable_full_run_date(ci_status, shippable)
      File "ansibullbot/triagers/plugins/needs_revision.py", line 666, in get_last_shippable_full_run_date
        rdata = shippable.get_run_data(to_text(runid), usecache=False)
      File "ansibullbot/utils/shippable_api.py", line 232, in get_run_data
        run_data = self._get_url(run_url, usecache=usecache)
      File "ansibullbot/utils/shippable_api.py", line 184, in _get_url
        self._write_cache_file(gzfile, [resp.status_code, jdata])
      File "ansibullbot/utils/shippable_api.py", line 141, in _write_cache_file
        f.write(json.dumps(data))
      File "/usr/lib/python3.7/gzip.py", line 260, in write
        data = memoryview(data)
    TypeError: memoryview: a bytes-like object is required, not 'str'

Manually tested using Python 2.7 and Python 3.7 using this command:

    ./triage_ansible.py --dry-run --pr 48991 --logfile log

PR #1121, opened 2018-12-10
Use ansibullbot._json_compat.json_dumps everywhere

pilou-:
See https://github.com/ansible/ansibullbot/pull/1117#discussion_r240239224. I am not sure why `json_dumps` wasn't already used everywhere ?

cc @webknjaz

### DIFF
--- a/ansibullbot/triagers/ansible.py
+++ b/ansibullbot/triagers/ansible.py
@@ -38,7 +38,7 @@ import pytz
 import requests
 import six
 
-from ansibullbot._json_compat import json_dump
+from ansibullbot._json_compat import json_dump, json_dumps
 from ansibullbot._text_compat import to_bytes, to_text
 
 import ansibullbot.constants as C
@@ -1545,7 +1545,7 @@ class AnsibleTriage(DefaultTriager):
         # cache the data in case of errors
         mdata = self.IM.migration_map.get(issue.html_url)
         with open(mfile, 'wb') as f:
-            f.write(json.dumps(mdata, indent=2))
+            f.write(json_dumps(mdata))
 
     def add_repomerge_comment(self, issue, actions, bp=u'repomerge'):
         '''Add the comment without closing'''
@@ -2450,4 +2450,4 @@ class AnsibleTriage(DefaultTriager):
         }
         resume_file = os.path.join(self.cachedir_base, u'resume.json')
         with open(resume_file, 'wb') as f:
-            f.write(json.dumps(data, indent=2))
+            f.write(json_dumps(data))
--- a/ansibullbot/utils/gh_gql_client.py
+++ b/ansibullbot/utils/gh_gql_client.py
@@ -4,7 +4,6 @@
 # https://developer.github.com/v4/guides/forming-calls/
 
 import jinja2
-import json
 import logging
 import requests
 from collections import defaultdict
@@ -13,6 +12,7 @@ from operator import itemgetter
 import six
 
 from tenacity import retry, wait_random, stop_after_attempt
+from ansibullbot._json_compat import json_dumps
 from ansibullbot._text_compat import to_bytes, to_text
 from ansibullbot.utils.receiver_client import post_to_receiver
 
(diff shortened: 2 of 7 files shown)


webknjaz:
I probably just missed some places.

webknjaz:
@abadger plz take a look as well

abadger:
I don't think this is quite correct and looking at the iterencode() method... it still looks wrong to me.

* For this PR, I think you need to use to_bytes around most of the json.dumps() calls.  I'll make a suggestion for one of the calls so that you can see what I mean

* For iterencode... why is the text transformation dependent on whether ensure_ascii is set?  IIRC, ensure_ascii means that non-ascii will be escaped in the output.  The input can contain non-ascii which should be text type in either case.

ansibullbot/triagers/ansible.py
```diff
@@ -1545,7 +1545,7 @@ def move_issue(self, issue):
         # cache the data in case of errors
         mdata = self.IM.migration_map.get(issue.html_url)
         with open(mfile, 'wb') as f:
-            f.write(json.dumps(mdata, indent=2))
+            f.write(json_dumps(mdata))
```
abadger:
```suggestion
            f.write(to_bytes(json_dumps(mdata), errors='surrogate_or_strict'))
```

The file is opened for writing in binary mode.  So you always want to write bytes to the file, not text.

Could you do this the other way and open the file in non-binary mode?  That won't work on python2 as open in python2 always expects bytes.  So you can get UnicodeError tracebacks if you write text to the filehandle on python2 when python2 implicitly tries to convert from text to bytes.

ansibullbot/utils/gh_gql_client.py
```diff
@@ -227,7 +227,7 @@ def get_summaries(self, owner, repo, otype='issues', last=None, first='first: 10
                 u'variables': u'{}',
                 u'operationName': None
             }
-            rr = requests.post(self.baseurl, headers=self.headers, data=json.dumps(payload))
+            rr = requests.post(self.baseurl, headers=self.headers, data=json_dumps(payload))
```
abadger: From this old bug: https://github.com/requests/requests/issues/3875 you may have to use to_bytes() with this as well.  I'm not certain, though.

ansibullbot/utils/shippable_api.py
```diff
@@ -138,7 +139,7 @@ def _load_cache_file(self, cfile):
 
     def _write_cache_file(self, cfile, data):
         with gzip.open(cfile, 'w') as f:
```
abadger: This should be:
```suggestion
        with gzip.open(cfile, 'wb') as f:
```

ansibullbot/wrappers/defaultwrapper.py
```diff
@@ -1016,7 +1017,7 @@ def log_ci_status(self, status_data):
                 jdata[turl][u'history'][ts] = sd[u'state']
 
         with open(logfile, 'w') as f:
```
abadger:
```suggestion
        with open(logfile, 'wb') as f:
```

ansibullbot/utils/gh_gql_client.py
webknjaz: Well, requests methods support `json=` arg so it's better just use it then:
```suggestion
            rr = requests.post(self.baseurl, headers=self.headers, json=payload)
```

ansibullbot/triagers/ansible.py
```diff
@@ -2450,4 +2450,4 @@ def set_resume(self, repo, number):
         }
         resume_file = os.path.join(self.cachedir_base, u'resume.json')
         with open(resume_file, 'wb') as f:
-            f.write(json.dumps(data, indent=2))
+            f.write(json_dumps(data))
```
webknjaz: same as above I guess

abadger: Yep, all the f.write's should do to_bytes() on the data.

ansibullbot/utils/gh_gql_client.py
abadger: Sounds good to me.

webknjaz:
@pilou- would you mind addressing the comments, please?

mkrizek:
I am going to go ahead and close this due to inactivity and because there is a plan to remove Python 2 support from the bot anyway.

Thanks @pilou- !

PR #1148, opened 2019-01-14
remove @bpennypacker

bpennypacker:
Removing myself as a maintainer


gundalow:
Hi,
The file you've edited is for tests.

https://github.com/ansible/ansible/blob/devel/.github/BOTMETA.yml is the file that's actually used. You may wish to add yourself to `ignored:` on the files or directories you are no longer wishing to be a maintainer.

Thanks for your help in the past.

bpennypacker:
Strange. I don't see myself OR any of the modules I commented listed in that file. So why then do I keep getting e-mails about commits?

PR #1160, opened 2019-02-03
Nullify yaml aliases to prevent pointer overload

jctanner:
Fixes https://github.com/ansible/ansibullbot/issues/1159


jctanner:
@dagwieers this look okay to you?

dagwieers:
@jctanner Thanks ! I assume this means my hunch was right. It probably has some performance impact on @ansibot since it uses more memory ?

Issue #1163, opened 2019-02-08
Ignore keywords/labels from example blocks or template sections


dagwieers:
So we currently have the problem that keywords or names are being gathered from the issue description, which is fine in general, but almost every issue is currently being identified with *lib/ansible/utils/parsing/yaml.py* because the description includes a ` ```yaml` block as part of the template.

In the same light, I would like @ansibot to not look for keywords in the ENVIRONMENT block of the description, so that a reported issue is not flagged as related to Windows or FreeBSD just because that is the environment where things may have been running. If the user suspects it to be related, this normally is already mentioned in the SUMMARY or title anyway.

jctanner:
Aha! I didn't realize someone had put "yaml" in that block. I can have the description parser start stripping that out. Thanks for the clue!

dagwieers:
I would be that person :-/

jctanner:
So yeah, the current code in extract_template_data is returning things like ...

```
(Epdb) pp tdict
{u'component name': u'vcenter_license',
 u'component_raw': u'vcenter_license\nor\nvmware_host\nADDITIONAL INFORMATION\nyaml',
 u'issue type': u'feature idea',
 u'summary': u'I was using vcenter_license to apply a license to ESXi hosts. Since the last commit on Jan 10, 2019, that module only supports licensing a vCenter. Add support for ESXi licensing'}
```

PR #1164, opened 2019-02-08
Use a simpler extraction method by default

jctanner:
Fixes https://github.com/ansible/ansibullbot/issues/1163


jctanner:
I will merge this on monday.

Issue #1169, opened 2019-02-15
AttributeError: 'NoneType' object has no attribute 'lower'


jctanner:
```
2019-02-15 09:08:10,538 DEBUG . /home/jtanner/.ansibullbot/cache/ansible.checkout/hacking/env-setup; ansible-doc -t module -F
2019-02-15 09:08:12,275 DEBUG no documentation found in /home/jtanner/.ansibullbot/cache/ansible.checkout/lib/ansible/modules/cloud/azure/_azure.py
2019-02-15 09:08:12,288 DEBUG no documentation found in /home/jtanner/.ansibullbot/cache/ansible.checkout/lib/ansible/modules/network/cumulus/_cl_interface_policy.py
2019-02-15 09:08:12,289 DEBUG no documentation found in /home/jtanner/.ansibullbot/cache/ansible.checkout/lib/ansible/modules/cloud/amazon/_ec2_vpc.py
2019-02-15 09:08:12,333 DEBUG no documentation found in /home/jtanner/.ansibullbot/cache/ansible.checkout/lib/ansible/modules/cloud/cloudstack/_cs_nic.py
2019-02-15 09:08:12,352 DEBUG no documentation found in /home/jtanner/.ansibullbot/cache/ansible.checkout/lib/ansible/modules/cloud/amazon/_ec2_remote_facts.py
2019-02-15 09:08:12,364 DEBUG no documentation found in /home/jtanner/.ansibullbot/cache/ansible.checkout/lib/ansible/modules/network/citrix/_netscaler.py
2019-02-15 09:08:12,379 DEBUG no documentation found in /home/jtanner/.ansibullbot/cache/ansible.checkout/lib/ansible/modules/windows/setup.ps1
2019-02-15 09:08:12,514 ERROR Uncaught exception
Traceback (most recent call last):
  File "./triage_ansible.py", line 45, in <module>
    main()
  File "./triage_ansible.py", line 41, in main
    AnsibleTriage().start()
  File "/home/jtanner/workspace/github/jctanner/ansibullbot/ansibullbot/triagers/ansible.py", line 242, in __init__
    email_cache=self.module_indexer.emails_cache
  File "/home/jtanner/workspace/github/jctanner/ansibullbot/ansibullbot/utils/component_tools.py", line 129, in __init__
    self.update()
  File "/home/jtanner/workspace/github/jctanner/ansibullbot/ansibullbot/utils/component_tools.py", line 135, in update
    self.index_files()
[...]

PR #1170, opened 2019-02-15
Refactor author extraction into something simpler+stabler

jctanner:
Fixes https://github.com/ansible/ansibullbot/issues/1169

### DIFF
--- a/ansibullbot/utils/component_tools.py
+++ b/ansibullbot/utils/component_tools.py
@@ -219,6 +219,15 @@ class AnsibleComponentMatcher(object):
             else:
                 _k = k
             ME = ModuleExtractor(os.path.join(checkoutdir, _k), email_cache=self.email_cache)
+
+            '''
+            try:
+                ME.authors
+            except AttributeError as e:
+                print(e)
+                import epdb; epdb.st()
+            '''
+
             if k not in self.BOTMETA[u'files']:
                 self.BOTMETA[u'files'][k] = {
                     u'deprecated': os.path.basename(k).startswith(u'_'),
--- a/ansibullbot/utils/component_tools.py
+++ b/ansibullbot/utils/component_tools.py
@@ -220,14 +220,6 @@ class AnsibleComponentMatcher(object):
                 _k = k
             ME = ModuleExtractor(os.path.join(checkoutdir, _k), email_cache=self.email_cache)
 
-            '''
-            try:
-                ME.authors
-            except AttributeError as e:
-                print(e)
-                import epdb; epdb.st()
-            '''
-
             if k not in self.BOTMETA[u'files']:
                 self.BOTMETA[u'files'][k] = {
                     u'deprecated': os.path.basename(k).startswith(u'_'),
--- a/ansibullbot/utils/component_tools.py
+++ b/ansibullbot/utils/component_tools.py
@@ -219,7 +219,6 @@ class AnsibleComponentMatcher(object):
             else:
                 _k = k
             ME = ModuleExtractor(os.path.join(checkoutdir, _k), email_cache=self.email_cache)
-
             if k not in self.BOTMETA[u'files']:
                 self.BOTMETA[u'files'][k] = {
                     u'deprecated': os.path.basename(k).startswith(u'_'),
--- a/ansibullbot/utils/extractors.py
+++ b/ansibullbot/utils/extractors.py
@@ -534,7 +534,7 @@ class ModuleExtractor(object):
         """Grep the authors out of the module docstrings"""
 
         # 2019-02-15
-        if 'author' self.docs or 'authors' in self.docs:
+        if 'author' in self.docs or 'authors' in self.docs:
             _authors = self.docs.get('author') or self.docs.get('authors')
             if _authors is None:
                 return []
@@ -551,7 +551,7 @@ class ModuleExtractor(object):
             return []
 
     def extract_github_id(self, author):
-        '''get github login(s) from a string'''
+        """Extract a set of github login(s) from a string."""
 
         # safegaurd against exceptions
         if author is None:
--- a/ansibullbot/utils/extractors.py
+++ b/ansibullbot/utils/extractors.py
@@ -496,36 +496,37 @@ class ModuleExtractor(object):
 
     @property
     def docs(self):
+        if self._DOCSTRING is not None:
+            return self._DOCSTRING
+
+        documentation = u''
+        inphase = False
+        lines = to_text(self.filedata).split(u'\n')
+        for line in lines:
+            if u'DOCUMENTATION' in line:
+                inphase = True
+                continue
+            if inphase and (line.strip().endswith((u"'''", u'"""'))):
+                #phase = None
+                break
+            if inphase:
+                documentation += line + u'\n'
+    
+        self._DOCUMENTATION_RAW = documentation
+
+        # some docstrings don't pass yaml validation with PyYAML >= 4.2
+        try:
+            self._DOCSTRING = yaml.load(self._DOCUMENTATION_RAW)
+        except yaml.parser.ParserError as e:
+            #logging.debug(e)
+            logging.warning('%s has non-yaml formatted docstrings' % self.filepath)
+        except yaml.scanner.ScannerError as e:
+            #logging.debug(e)
+            logging.warning('%s has non-yaml formatted docstrings' % self.filepath)
+
+        # always cast to a dict for easier handling later
         if self._DOCSTRING is None:
-
-            documentation = u''
-            inphase = False
-            lines = to_text(self.filedata).split(u'\n')
-            for line in lines:
-                if u'DOCUMENTATION' in line:
-                    inphase = True
-                    continue
-                if inphase and (line.strip().endswith((u"'''", u'"""'))):
-                    #phase = None
-                    break
-                if inphase:
-                    documentation += line + u'\n'
-        
-            self._DOCUMENTATION_RAW = documentation
-
-            # some docstrings don't pass yaml validation with PyYAML >= 4.2
-            try:
-                self._DOCSTRING = yaml.load(self._DOCUMENTATION_RAW)
-            except yaml.parser.ParserError as e:
-                #logging.debug(e)
-                logging.warning('%s has non-yaml formatted docstrings' % self.filepath)
-            except yaml.scanner.ScannerError as e:
-                #logging.debug(e)
-                logging.warning('%s has non-yaml formatted docstrings' % self.filepath)
-
-            # always cast to a dict for easier handling later
-            if self._DOCSTRING is None:
-                self._DOCSTRING = {}
+            self._DOCSTRING = {}
 
         return self._DOCSTRING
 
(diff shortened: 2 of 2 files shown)


ansibullbot/utils/component_tools.py
```diff
@@ -219,6 +219,15 @@ def index_files(self):
             else:
                 _k = k
             ME = ModuleExtractor(os.path.join(checkoutdir, _k), email_cache=self.email_cache)
+
+            '''
```
webknjaz: This should probably not be committed

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
-                inphase = True
-                continue
-            if inphase and (line.strip().endswith((u"'''", u'"""'))):
-                #phase = None
-                break
-            if inphase:
-                documentation += line + u'\n'
+            try:
+                self._DOCSTRING = yaml.load(self._DOCUMENTATION_RAW)
+            except:
```
webknjaz: proper exception plz

ansibullbot/utils/component_tools.py
```diff
@@ -219,6 +219,7 @@ def index_files(self):
             else:
                 _k = k
             ME = ModuleExtractor(os.path.join(checkoutdir, _k), email_cache=self.email_cache)
+
```
webknjaz:
```suggestion
```

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
             self._METADATA = self.get_module_metadata()
         return self._METADATA
 
-    def get_module_authors(self):
-        """Grep the authors out of the module docstrings"""
+    @property
+    def docs(self):
+        if self._DOCSTRING is None:
```
webknjaz: Pls reverse this check to be in a "guard expression" style. This will allow dedenting huge chunk of code. Flat src is better readable.
```suggestion
        if self._DOCSTRING is not None:
            return self._DOCSTRING
```

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
-            print(e)
-            return []
+    def get_module_authors(self):
+        """Grep the authors out of the module docstrings"""
 
-        # quit early if the yaml was not valid
-        if not ydata:
-            return []
+        # 2019-02-15
+        if 'author' or 'authors' in self.docs:
```
webknjaz: `if 'author'` is always `True`. Did you mean
```suggestion
        if 'author' in self.docs or 'authors' in self.docs:
```

webknjaz: or maybe
```suggestion
        if {'author', 'authors'} & self.docs.keys():
```

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
+            _authors = self.docs.get('author') or self.docs.get('authors')
```
webknjaz: Maybe you could normalize this or put into a property?

webknjaz: Oh, and which one takes priority? singular or plural?

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
-            github_ids = self.extract_github_id(author)
-            if github_ids:
-                authors.extend(github_ids)
+    def extract_github_id(self, author):
+        '''get github login(s) from a string'''
 
-        return authors
+        # safegaurd against exceptions
+        if author is None:
+            return []
```
webknjaz: This is a different data type from `set` which is returned by default which may crash `.union()` call.

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
-        if type(ydata[u'author']) != list:
-            ydata[u'author'] = [ydata[u'author']]
-
-        authors = []
-        for author in ydata[u'author']:
-            github_ids = self.extract_github_id(author)
-            if github_ids:
-                authors.extend(github_ids)
+    def extract_github_id(self, author):
+        '''get github login(s) from a string'''
```
webknjaz:
```suggestion
        """Extract a set of github login(s) from a string."""
```

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
+            _authors = self.docs.get('author') or self.docs.get('authors')
+            if _authors is None:
+                return []
+            if not isinstance(_authors, list):
+                _authors = [_authors]
+            logins = set()
+            for author in _authors:
+                _logins = self.extract_github_id(author)
+                if _logins:
+                    logins = logins.union(_logins)
```
webknjaz: self.extract_github_id() returns a possible mixed type (list or set). list will crash this.

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
+        if 'author' or 'authors' in self.docs:
+            _authors = self.docs.get('author') or self.docs.get('authors')
+            if _authors is None:
+                return []
+            if not isinstance(_authors, list):
+                _authors = [_authors]
+            logins = set()
+            for author in _authors:
+                _logins = self.extract_github_id(author)
+                if _logins:
```
webknjaz: in case of set, it's fine to skip this check.

ansibullbot/utils/extractors.py
webknjaz: Oh... the list is returned but `get_module_authors()` expects set

ansibullbot/utils/extractors.py
webknjaz: oh, it's a different var... Anyway, that if-clause is unnecessary

webknjaz: maybe you could do this via pipe operator?
```suggestion
                    logins = logins | _logins
```

ansibullbot/utils/extractors.py
```diff
@@ -493,77 +494,63 @@ def metadata(self):
...
-            return []
+        # 2019-02-15
+        if 'author' or 'authors' in self.docs:
+            _authors = self.docs.get('author') or self.docs.get('authors')
+            if _authors is None:
+                return []
+            if not isinstance(_authors, list):
+                _authors = [_authors]
+            logins = set()
+            for author in _authors:
```
webknjaz: Now I see that you want just
```suggestion
            return list(set(self.extract_github_id(author) for author in _authors))
```

ansibullbot/utils/extractors.py
```diff
@@ -513,10 +513,15 @@ def docs(self):
...
             self._DOCUMENTATION_RAW = documentation
 
+            # some docstrings don't pass yaml validation with PyYAML >= 4.2
             try:
                 self._DOCSTRING = yaml.load(self._DOCUMENTATION_RAW)
-            except:
-                self._DOCSTRING = {}
+            except yaml.parser.ParserError as e:
+                #logging.debug(e)
+                logging.warning('%s has non-yaml formatted docstrings' % self.filepath)
```
webknjaz: Since this block is just copy-paste, you can put both cases into one except section

ansibullbot/utils/extractors.py
```diff
@@ -496,36 +496,37 @@ def metadata(self):
 
     @property
     def docs(self):
-        if self._DOCSTRING is None:
+        if self._DOCSTRING is not None:
```
webknjaz: FTR you could use smth like `@lru_cache(maxsize=1)`

ansibullbot/utils/extractors.py
jctanner: TypeError: unhashable type: 'list'

ansibullbot/utils/extractors.py
```diff
@@ -509,61 +511,52 @@ def get_module_authors(self):
...
 
-        # quit if the key was not found
-        if u'author' not in ydata:
-            return []
+        return self._DOCSTRING
 
-        if type(ydata[u'author']) != list:
-            ydata[u'author'] = [ydata[u'author']]
+    def get_module_authors(self):
+        """Grep the authors out of the module docstrings"""
```
webknjaz: Docstring title line should be a sentence ending with a period.
```suggestion
        """Grep the authors out of the module docstrings."""
```

ansibullbot/utils/extractors.py
```diff
@@ -509,61 +511,52 @@ def get_module_authors(self):
...
 
-        authors = []
-        for author in ydata[u'author']:
-            github_ids = self.extract_github_id(author)
-            if github_ids:
-                authors.extend(github_ids)
+        # 2019-02-15
+        if 'author' self.docs or 'authors' in self.docs:
```
webknjaz: This check could be simplified:
```suggestion
        _authors = self.docs.get('author') or self.docs.get('authors')
        if not _authors:
            return []

        # rest of code is unindented, extra similar check is not needed
```

ansibullbot/utils/extractors.py
```diff
@@ -509,61 +511,52 @@ def get_module_authors(self):
...
+    def get_module_authors(self):
+        """Grep the authors out of the module docstrings"""
 
-        authors = []
-        for author in ydata[u'author']:
-            github_ids = self.extract_github_id(author)
-            if github_ids:
-                authors.extend(github_ids)
+        # 2019-02-15
+        if 'author' in self.docs or 'authors' in self.docs:
```
webknjaz: This check could be simplified:
```suggestion
        _authors = self.docs.get('author') or self.docs.get('authors')
        if not _authors:
            return []

        # rest of code is unindented, extra similar check is not needed
```

ansibullbot/utils/extractors.py
jctanner: I have no idea. this is a thing the community did without my involvement.

ansibullbot/utils/extractors.py
```diff
@@ -493,12 +494,13 @@ def metadata(self):
...
 
-    def get_module_authors(self):
-        """Grep the authors out of the module docstrings"""
+    @property
+    def docs(self):
+        if self._DOCSTRING is not None:
+            return self._DOCSTRING
 
         documentation = u''
         inphase = False
```
webknjaz: plz return the removed line here:
```suggestion
         inphase = False

```

ansibullbot/utils/extractors.py
webknjaz: oh why?

webknjaz: oh it should probably be
```suggestion
            return list(set(a for author in _authors for a in self.extract_github_id(author)))
```

ansibullbot/utils/extractors.py
webknjaz: You can just use `author`. `authors` is forbidden now and is eliminated by current schema validator: https://github.com/ansible/ansibullbot/pull/1050

PR #1190, opened 2019-05-08
Add a reporting function for metadata

jctanner:


ansible_support_report.py
```diff
@@ -0,0 +1,33 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import print_function
+
+import logging
+import sys
+
+from ansibullbot.reports.ansible import AnsibleSupportReport
+from ansibullbot.utils.sentry import initialize_sentry
```
webknjaz: @jctanner you probably wanted to actually call this...

PR #1199, opened 2019-05-29
Process !component even if component is empty

mkrizek:
https://github.com/ansible/ansible/issues/56873#issuecomment-495392787


mkrizek:
@jctanner do you recall the logic behind that code?

PR #1233, opened 2019-07-15
use logins for merge checking

### LINKED ISSUE #1231: Automerge is broken
There is a bug about automerge.
It does not work well as expected after PR was labeled 'automerge'.
Such as ansible/ansible#59071

jctanner:
fixes #1231

### DIFF
--- a/ansibullbot/wrappers/defaultwrapper.py
+++ b/ansibullbot/wrappers/defaultwrapper.py
@@ -94,6 +94,7 @@ class DefaultWrapper(object):
         self._assignees = False
         self._comments = False
         self._committer_emails = False
+        self._committer_logins = False
         self._commits = False
         self._events = False
         self._history = False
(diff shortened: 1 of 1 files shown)


ansibullbot/wrappers/defaultwrapper.py
```diff
@@ -94,6 +94,7 @@ def __init__(self, github=None, repo=None, issue=None, cachedir=None, file_index
         self._assignees = False
         self._comments = False
         self._committer_emails = False
+        self._committer_logins = False
```
webknjaz: Putting both bool and list into the same var is probably not the best idea. How about
```suggestion
        self._committer_logins = set()
```

jctanner: It repeats the same pattern as all the other class vars, but I guess it could be switched.

jctanner: I'll file a bug to fix that across the board later.

PR #1293, opened 2020-02-28
Gundalow script updates

jctanner:


scripts/generate_collection_botmeta.py
```diff
@@ -0,0 +1,331 @@
...
+from ansibullbot.utils.logs import set_logger
+from ansibullbot.utils.botmeta import botmeta_list
+
+set_logger(debug=True)
+
+
+GITHUB_TOKEN = os.getenv('GITHUB_TOKEN')
+if GITHUB_TOKEN is None:
+    GITHUB_TOKEN = C.DEFAULT_GITHUB_TOKEN
+assert GITHUB_TOKEN, "GITHUB_TOKEN env var must be set to an oauth token with repo:read from https://github.com/settings/"
```
webknjaz: FYI `assert`s are stripped off and aren't evaluated in runtime when `python` is run in optimized mode.

gundalow:
CI is failing. Though maybe as this will only be run next week, we don't need to have this script committed to the main repo.

Are you happy with the script as is?

jctanner:
It works well enough as is from this feature branch. When I get some time, i'll fix the unit tests.

PR #1295, opened 2020-03-24
post-NWO cleanup

jctanner:


ansibullbot/parsers/botmetadata.py
```diff
@@ -148,16 +149,35 @@ def _propagate(files, top, child, field, multivalued=True):
...
-            for file1, file2 in itertools.combinations(files.keys(), 2):
-                # Python 2.7 doesn't provide os.path.commonpath
-                common = os.path.commonprefix([file1, file2])
-                top = min(file1, file2)
-                child = max(file1, file2)
 
-                top_components = top.split(u'/')
-                child_components = child.split(u'/')
+            iterfiles = {}
+            filenames = sorted(list(files.keys()))
```
abadger: You don't need list here.  `sorted()` will produce a list from an iterable.

ansibullbot/parsers/botmetadata.py
```diff
@@ -148,16 +149,35 @@ def _propagate(files, top, child, field, multivalued=True):
...
+            parent = None
```
abadger: This isn't needed because parent is only used inside of the loop where it's set.

ansibullbot/parsers/botmetadata.py
```diff
@@ -148,16 +149,35 @@ def _propagate(files, top, child, field, multivalued=True):
...
+
+            for file1, files2 in iterfiles.items():
+                for file2 in files2:
+                    # Python 2.7 doesn't provide os.path.commonpath
+                    common = os.path.commonprefix([file1, file2])
+                    top = min(file1, file2)
+                    child = max(file1, file2)
+
+                    top_components = top.split(u'/')
+                    child_components = child.split(u'/')
```
abadger: I think you can get rid of the common, top, child, top_components, and child_components setting now.  If I'm reading it right, these are all used to determine whether and which filename is the parent and which is the child.  That's now done when the information is put into iterfiles.  file1 == top and file2 == child.

ansibullbot/parsers/botmetadata.py
abadger: It may be faster to do the uniqueness check here instead of on line 163:

```suggestion
            filenames = sorted(frozenset(files))
```

It creates a frozenset temporarily which may decrease speed but then it's able to check for uniqueness in C code (the frozenset) instead of inside of a python loop (the if continue on line 163 and 164) which would improve speed.

ansibullbot/parsers/botmetadata.py
```diff
@@ -148,16 +149,35 @@ def _propagate(files, top, child, field, multivalued=True):
...
 
-                top_components = top.split(u'/')
-                child_components = child.split(u'/')
+            iterfiles = {}
+            filenames = sorted(list(files.keys()))
+            parent = None
+            for idp,parent in enumerate(filenames):
+
+                if parent not in iterfiles:
+                    iterfiles[parent] = []
```
abadger: You can make iterfiles a `defaultdict`instead of using this idiom.

`
-iterfiles = {}
+iterfiles = defaultdict(list)
`
on line 153

abadger:
This is a bit outside the realm of this PR but it would be great if you could unnest these functions and put them at the toplevel instead.  They're algorithmic in nature so it is natural to write unittests to test all the permutations of data that should go into them.  But right now, they're nested inside of other functions which are not purely algorithmic (the other functions depend on external resources like the filesystem) so testing all of the possibilities becomes a much bigger chore.

ansibullbot/triagers/plugins/component_matching.py
```diff
@@ -142,6 +145,14 @@ def get_component_match_facts(iw, component_matcher, valid_labels):
                 if u'/plugins/' in x:
                     cmeta[u'is_new_plugin'] = True
 
+    # is it a colleciton?
+    if [x for x in CM_MATCHES if x[u'repo_filename'].startswith(u'collection:')]:
+        cmeta[u'is_collection'] = True
+        cmeta[u'component_collection'] = []
+        cmeta[u'component_support'] = [u'community']
+        for comp in [x for x in CM_MATCHES if x[u'repo_filename'].startswith(u'collection:')]:
```
webknjaz: Using a generator expression will prevent the creation of an unnecessary in-memory list:
```suggestion
        for comp in (x for x in CM_MATCHES if x[u'repo_filename'].startswith(u'collection:')):
```

ansibullbot/triagers/plugins/component_matching.py
```diff
@@ -142,6 +145,14 @@ def get_component_match_facts(iw, component_matcher, valid_labels):
                 if u'/plugins/' in x:
                     cmeta[u'is_new_plugin'] = True
 
+    # is it a colleciton?
+    if [x for x in CM_MATCHES if x[u'repo_filename'].startswith(u'collection:')]:
```
webknjaz: Better avoid creating in-mem lists. This will also save time by stopping the iteration early:
```suggestion
    is_collection = any(x[u'repo_filename'].startswith(u'collection:') for x in CM_MATCHES)
    if is_collection:
```

webknjaz: Alternatively, you could save this list in a var since you use it later in code. So that you'd not repeat this twice:
```suggestion
    collection_components = [x for x in CM_MATCHES if x[u'repo_filename'].startswith(u'collection:')]
    if collection_components:
```
Then, you could reuse this var on line 153

ansibullbot/triagers/plugins/component_matching.py
```diff
@@ -142,6 +145,14 @@ def get_component_match_facts(iw, component_matcher, valid_labels):
...
+        cmeta[u'is_collection'] = True
+        cmeta[u'component_collection'] = []
+        cmeta[u'component_support'] = [u'community']
+        for comp in [x for x in CM_MATCHES if x[u'repo_filename'].startswith(u'collection:')]:
+            cmeta[u'component_collection'].append(x[u'repo_filename'].replace(u'collection:', ''))
```
webknjaz:
```suggestion
            cmeta[u'component_collection'].append(x[u'repo_filename'].replace(u'collection:', u''))
```

ansibullbot/utils/component_tools.py
```diff
@@ -472,6 +501,27 @@ def _match_component(self, title, body, component):
...
 
+    def search_by_galaxy(self, component):
+        '''Is this a file belonging to a collection?'''
+
+        matches = []
+
+        candidates = []
+        for key in self.GALAXY_FILES.keys():
+            if (component in key or key == component) and key.startswith('plugins'):
+                logging.info(u'matched %s to %s:%s' % (component, key, self.GALAXY_FILES[key]))
```
webknjaz:
```suggestion
        for key, val in self.GALAXY_FILES.items():
            if (component in key or key == component) and key.startswith('plugins'):
                logging.info(u'matched %s to %s:%s' % (component, key, val))
```

ansibullbot/utils/component_tools.py
```diff
@@ -472,6 +501,27 @@ def _match_component(self, title, body, component):
...
+
+        # FIXME: need some better validation of the candidates ...
+
+        if candidates:
+            for cn in candidates:
+                for fqcn in self.GALAXY_FILES[cn]:
+                    matches.append('collection:%s' % fqcn)
+            matches = sorted(set(matches))
+
+        return matches
```
webknjaz:
```suggestion
        if not candidates:
        	return []

        for cn in candidates:
            for fqcn in self.GALAXY_FILES[cn]:
                matches.append('collection:%s' % fqcn)

        return sorted(set(matches))
```

Issue #1161, opened 2019-02-05
Feature: Add commands for rebuilding CI


dagwieers:
In those cases where the CI tests FAILED or were deemed UNSTABLE it would be nice if the contributor does not have to ask on #ansible-devel to rebuild the failed CI tests, or has to close and reopen the PR (which causes *all* tests to be re-run). But instead has the option to do `rebuild_ci`.

@ansibot does not have to report back (or add a comment) it simply needs to reschedule a CI rebuild job.

mattclay:
Related issue: https://github.com/ansible/ansibullbot/issues/1105

Yes, Shippable does have an API to restart failed jobs.

gundalow:
@mattclay Would we want:
`rerun_ci`
`rerun_failed_ci` - run a subset

mattclay:
@gundalow We probably want to make those commands Shippable specific, to distinguish them from any commands that may affect Zuul. Perhaps something like:

`rerun_shippable` - re-run only failed Shippable jobs (ignored if no failures)
`rerun_shippable_all` - re-run all Shippable jobs

Most of the time re-running all jobs shouldn't be needed, so it's better to be explicit about that.

gundalow:
Need to also ensure that the Bot re-reads Shippable status

Who should be able to use this command?
* PR Author
* Anyone with commit?
* maintainer for 1+ of the files
* Any maintainer?

mattclay:
I'd start by allowing both the PR author and anyone with commit, since they're already capable of pushing new commits and closing/reopening to trigger a full CI run.

pabelanger:
so, I am a fan of recheck if we are painting the bike shed :)

For zuul.ansible.com, this won't be an issue, we can match on arbitrary comment strings.

samdoran:
I'm a fan of `recheck` being the command.

mattclay:
We should be able to re-run tests on CI systems independently, so using `recheck` for Zuul would be fine, but we'd want something different for Shippable.

felixfontein:
I could really use this!

agaffney:
Now that external contributors no longer have commit access to `ansible/ansible` to restart failed CI jobs, this would be very useful

jctanner:
initial code supplied by https://github.com/ansible/ansibullbot/pull/1299

PR #1299, opened 2020-03-31
add /rebuild command for all users

jctanner:
https://github.com/ansible/ansibullbot/issues/1161


jctanner:
I failed to realize the request for this command was intended to only rebuild the failed jobs instead all jobs. I'll keep digging via another PR.

PR #1343, opened 2020-05-18
Fix flatten map

mkrizek:


mkrizek:
https://github.com/ansible/ansible/issues/69034

PR #1408, opened 2020-08-01
Don't add component_name if we didn't found one.

### LINKED ISSUE #1301: do not use component template if component is None
https://github.com/ansible/ansible/issues/68619

bmillemathias:
If the expected list of components found is empty, do nothing.
I nonetheless modified also the template so when no component is found
we write it explicitely instead of return 'None'

Fixes #1301


mkrizek:
Shouldn't we still post the comment suggesting to use the component command when the bot is not able to figure out the component?

I think it makes sense to omit the comment only if the whole component section is missing from the template.

mkrizek:
@bmillemathias are you still interested on working on this PR?

mkrizek:
Closing due to inactivity.

PR #1410, opened 2020-08-07
Add bot_closed label when auto-closed due to needs_collection_redirect

mkrizek:


mkrizek:
cc @sivel (can't be added as a reviewer via GH UI)

mkrizek:
@gundalow @jimi-c @sivel With this patch, the bot would add `bot_closed` label only for issues labeled `needs_collection_redirect`. Alternatively we can add that label to all issues that bot closes and then to find and review issues auto-closed due to collection redirect we would use both `bot_closed` and `needs_collection_redirect` labels.

Also, I would not turn on the auto-close feature for collection issues on Friday :-)

sivel:
I'm good with this as is.

Issue #1197, opened 2019-05-27
Fix `.gitmodules` to point to the right versions of submodule states


webknjaz:
Currently, this config shows to an ephemeral branch. It's better to lock things to certain permanent refs like sha or tag.

Issue #1134, opened 2018-12-18
caddy and ansibullbot roles have unmet group dependency


jctanner:
```
    default: TASK [mongodb : Include distribution specific variables] ***********************
    default: ok: [ansibullbot.eng.ansible.com] => (item=/vagrant/playbooks/roles/mongodb/vars/RedHat.yml) => {"ansible_facts": {"mongodb_config_file": "/etc/mongod.conf", "mongodb_data_dir": "/var/lib/mongo", "mongodb_group": "mongod", "mongodb_log_dir": "/var/log/mongodb", "mongodb_log_file": "{{ mongodb_log_dir }}/mongod.log", "mongodb_packages": ["mongodb-org"], "mongodb_pid_file": "/var/run/mongodb/mongod.pid", "mongodb_selinux_ports": "{{ mongodb_port ~ ',' ~ mongodb_shardsvr_port ~ ',' ~ mongodb_configsvr_port ~ ',' ~ mongodb_webstatus_port }}", "mongodb_service_name": "mongod", "mongodb_user": "mongod"}, "ansible_included_var_files": ["/vagrant/playbooks/roles/mongodb/vars/RedHat.yml"], "changed": false, "item": "/vagrant/playbooks/roles/mongodb/vars/RedHat.yml"}
    default: 
    default: TASK [caddy : Include distribution specific variables] *************************
    default: ok: [ansibullbot.eng.ansible.com] => (item=/vagrant/playbooks/roles/caddy/vars/RedHat.yml) => {"ansible_facts": {"caddy_packages": ["caddy"], "caddy_ssl_path": "/etc/pki/tls/caddy"}, "ansible_included_var_files": ["/vagrant/playbooks/roles/caddy/vars/RedHat.yml"], "changed": false, "item": "/vagrant/playbooks/roles/caddy/vars/RedHat.yml"}                                                                                                  
    default: 
    default: TASK [caddy : include] *********************************************************
    default: included: /vagrant/playbooks/roles/caddy/tasks/install-RedHat.yml for ansibullbot.eng.ansible.com
    default: 
    default: TASK [ansibullbot : MONITORING | Add caddy user account to ansibot group] ******
    default: fatal: [ansibullbot.eng.ansible.com]: FAILED! => {"changed": false, "msg": "Group ansibot does not exist"}                                                                                                          
    default:    to retry, use: --limit @/vagrant/playbooks/setup-ansibullbot.retry
    default: 
[...]

PR #1444, opened 2020-10-08
Update bot deployment

samdoran:
- CentOS 8
- Python 3.8 in a virtualenv
- updated roles


.gitmodules
```diff
@@ -25,3 +25,6 @@
 [submodule "playbooks/roles/caddy"]
 	path = playbooks/roles/caddy
 	url = https://github.com/samdoran/ansible-role-caddy.git
+[submodule "playbooks/roles/authorized_keys"]
```
mkrizek, line 28: Since you are changing this file, I remembered there is an issue https://github.com/ansible/ansibullbot/issues/1197, do we want to fix that (not necessarily in this PR)? If not, feel free to close.

mkrizek:
Is https://github.com/ansible/ansibullbot/issues/1134 something that is still an issue? As my previous comment not something that needs to happen in this PR (or at all), just that since you are updating the deployment I thought I'd mention that.

.gitmodules
samdoran: Yup, I will look into it.

PR #1471, opened 2020-12-01
Switch to Azure Pipelines

samdoran:
- Update main `ansibullbot.cfg` config template in the role.
- Update host_vars to use Azure pipelines instead of Shippable


samdoran:
Waiting on the creation of the pipeline in order to get the correct project ID. Once we have that, we can set it correctly in this PR and merge it.

mkrizek (approved):
👍

PR #1488, opened 2021-01-19
Fix Ansibullbot status page

samdoran:
- Get the cgi script running properly
- Fix Caddy config to allow the cgi script to run


samdoran:
Despite several hours of work, I'm not able to get this working then SELinux is enabled. 😞 

I tried compiling a custom module and enabling the `domain_can_mmap_files` SELinux boolean, but I'm still missing something.

The script does execute correctly from the `/status` page now, but only when SELinux is disabled.

PR #1500, opened 2021-02-02
Debug missing check_runs

mkrizek:


ansibullbot/wrappers/defaultwrapper.py
```diff
@@ -505,6 +505,9 @@ def pullrequest_check_runs(self):
             url = 'https://api.github.com/repos/%s/commits/%s/check-runs' % (self.repo_full_name, self.pullrequest.head.sha)
             self._pullrequest_check_runs = []
             for resp_data in self.github.get_request_gen(url):
+                if 'check_runs' not in resp_data:
```
webknjaz, line 508: `resp_data` was `{message: 'Server Error}` at the time of error:
https://sentry.io/organizations/red-hat-ansibullbot/issues/2164302248/?project=1336908&referrer=slack

In fact, I just opened https://api.github.com/repos/ansible/ansible/commits/bce0165/check-runs in my browser and it returns the same response.

mkrizek: > In fact, I just opened https://api.github.com/repos/ansible/ansible/commits/bce0165/check-runs in my browser and it returns the same response in the body and HTTP 502 as a status.

Where did that link come from? It failed with 502 at https://api.github.com/repositories/3638964/commits/2b190305a526d64269b314361fdb2fcf2e31f972/check-runs?page=4.

webknjaz: > Where did that link come from?

I just constructed the URL with the latest commit from devel manually as a test.

Issue #1047, opened 2018-09-21
classify/label PRs that are docs only (not other files)


jctanner:
anything in docsite only
anything in examples
anything that doesn't change python code
anything that only changes the docstrings in module

jctanner:
cc @acozine

acozine:
Thanks, this would help the docs team identify PRs that we can merge without code review.

mkrizek:
Fixed in #1505

PR #1505, opened 2021-02-06
Classify/Label PRs That Are Docs Only

sommersoft:
This is an approach to address #1047.

Working with the currently available labels in ansible/ansible:
 - Will add the ~~`docsite`~~ `docs_only` label to PRs if ALL changes are limited to "docs only" logic.
 - Will remove a ~~`docsite`~~ `docs_only` label if there are changes outside of "docs only", unless the label was applied by a human.
 - Will not override/remove any other labels that the triager determines are applicable.

I couldn't determine with any certainty how ansibullbot adds the `docsite` label (via `utils/component_tools.py`, etc.); it seemed mostly random looking at various PRs and the component matching routines.

This is written quite narrowly at the moment, but could possibly be expanded for use in other issues (notification, standards enforcement, etc).

Very small sample set (PRs are now merged):
```shell
##################################################################
# ansible/ansible PR: 73287                                      #
# Changes: 1. The `DOCUMENTATION` variable in postgresql_user.py #
##################################################################
$ ./triage_ansible.py --dry-run --debug --verbose --pr 73287 --logfile ../ansibot_log/log.txt --ignore_galaxy

2021-01-25 08:17:13,383 INFO 4670 ansible.py:__init__:228 starting bot
....
2021-01-25 08:17:51,694 INFO 4670 ansible.py:run:524 url: https://github.com/ansible/ansible/pull/73287
2021-01-25 08:17:51,694 INFO 4670 ansible.py:run:525 title: [2.9] postgresql_user: add clarification about no_password_changes option
2021-01-25 08:17:51,695 INFO 4670 ansible.py:run:528 component[f]: lib/ansible/modules/database/postgresql/postgresql_user.py
2021-01-25 08:17:51,695 INFO 4670 ansible.py:run:541 needs_revision
{'assign': [],
 'cancel_ci': False,
 'cancel_ci_branch': False,
 'close': False,
 'close_migrated': False,
 'comments': [],
 'merge': False,
 'newlabel': ['docsite', 'needs_ci', 'needs_revision'], # <--- docsite label added by get_docs_facts()
 'open': False,
 'rebuild': False,
 'rebuild_failed': False,
 'unassign': [],
 'uncomment': [],
 'unlabel': ['collection:community.postgresql', 'community_review']}
Dry-run specified, skipping execution of actions


##################################################################
# ansible/ansible PR: 73334                                      #
# Changes: 1. Added a changelog fragment                         #
#          2. The `DOCUMENTATION` & `RETURNS` variables, and     #
#             several logic changes in apt_key.py                #
#          3. Changes to integration tests                       #
##################################################################
$ ./triage_ansible.py --dry-run --debug --verbose --pr 73334 --logfile ../ansibot_log/log.txt --ignore_galaxy

2021-01-25 08:22:26,017 INFO 5551 ansible.py:__init__:228 starting bot
....
2021-01-25 08:23:17,367 INFO 5551 ansible.py:run:524 url: https://github.com/ansible/ansible/pull/73334
2021-01-25 08:23:17,372 INFO 5551 ansible.py:run:525 title: change detection and check mode fixes for apt_key
2021-01-25 08:23:17,372 INFO 5551 ansible.py:run:528 component[f]: changelogs/fragments/apt_key_fixes.yml
2021-01-25 08:23:17,372 INFO 5551 ansible.py:run:528 component[f]: lib/ansible/modules/apt_key.py
2021-01-25 08:23:17,372 INFO 5551 ansible.py:run:541 needs_revision
{'assign': [],
 'cancel_ci': False,
 'cancel_ci_branch': False,
 'close': False,
 'close_migrated': False,
 'comments': [],
 'merge': False,
 'newlabel': ['needs_ci'], # <--- docsite label not added by get_docs_facts()
 'open': False,
 'rebuild': False,
 'rebuild_failed': False,
 'unassign': [],
 'uncomment': [],
 'unlabel': []}
Dry-run specified, skipping execution of actions
```

The attached logfile contains more examples (50 open PRs at time of running).
[log_new_docsite.txt](https://github.com/ansible/ansibullbot/files/5937005/log_new_docsite.txt)

### Open questions:
1. Are the directory matching patterns the desired ones?
2. The AST parsing and comparison functions ignore module-level/non-class functions. Is this correct?
3. ~~How are Galaxy components affected? I didn't see any activity from this addition when running my tests with Galaxy included, but I'm not certain there are no effects.~~
4. ~~Is a test desired for this?~~
    - Unit test created for `get_docs_facts`.
5. ~~Is `docsite` the appropriate label to assign, or should a new label be created on `ansible/ansible`?~~
    -   `docs_only` label created, and will be used.


ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+                        break
+                    # If a python file, check if the changes are only to
+                    # docstrings
+                    else:
+                        docs_only = _check_py_changes(
+                                changed_file.file_content,
+                                changed_file.patch
+                        )
+
+        if not docs_only:
```
mkrizek, line 302: Seems like this is incorrectly indented as it should be checked per `changed_file`.

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+            dfacts["is_docs_only"] = "docsite" in iw.labels
+            return dfacts
+
+        for changed_file in commit_files:
+            if isinstance(changed_file, dict):
+                changed_file = CommitFile(changed_file)
+
+            if _is_docs_path(changed_file.filename):
+                continue
+            else:
```
mkrizek, line 281: All these `else` are not really needed since there are `continue` or `break` used.

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+        affect docs-related files/content.
+    """
+    dfacts = {
+        "is_docs_only": False
+    }
+
+    if not iw.is_pullrequest():
+        return dfacts
+
+    docs_only = True
```
mkrizek, line 265: Should we rather start with `docs_only` being set to `False` to prevent false positives in case say `iw.commits` misbehaves and returns an empty list? I have not looked into how likely that is though.

mkrizek:
Thanks for working on this.

I was a bit confused at first because the description mentions adding `docsite` label but in fact this should add a new label something like `docs_only` instead of re-using existing `docsite` label that is used for pull requests that change content of `docs/docstite/` directory (documentation website).

I haven't looked at the code at detail yet, just quicky skimmed through `get_docs_facts` and left some comments there.

> Are the directory matching patterns the desired ones?

Maybe use `docs/` instead of just `docs/docsite`?

> The AST parsing and comparison functions ignore module-level/non-class functions. Is this correct?

I am not sure. Why ignore those?

> How are Galaxy components affected? I didn't see any activity from this addition when running my tests with Galaxy included, but I'm not certain there are no effects.

Anything that is in galaxy should be auto-closed anyway so we don't have to take them into account.

> Is a test desired for this?

Yes, please.

cc'ing @acozine @samccann  for any input from the docs team.

sommersoft:
@mkrizek, thanks for the review. Your suggestions (along with LGTM's) did highlight that I should've polished a little more before submitting. Some of those are left-over after munging my standalone prototype into ansibullbot.

> ...the description mentions adding `docsite` label but in fact this should add a new label something like `docs_only` instead of re-using existing `docsite`...

I agree that `docs_only`, or something similar and separate, would be better. I just wanted to work with a label that existed in the ansible/ansible repo to get this started. I actually meant to add this to the **Open Questions** (will edit/expand after this message).

> The AST parsing and comparison functions ignore module-level/non-class functions. Is this correct?
>> I am not sure. Why ignore those?

My thinking was that most of Ansible's documented API is class based, making module-level functions a rarity for the docs. However, Ansible is a **large** project, so this could entirely be false; especially for edge cases. It wouldn't take much to add them. Would just need to rearrange the data structure and add another conditional.

> Is a test desired for this?
>> Yes, please.

After codecov ran, I figured that would be the case. 😄

ansibullbot/triagers/plugins/docs_info.py
sommersoft: This is to break out of the parent loop that is iterating over each commit. This is basically implementing a "fail fast" scenario, since there was a non-docs change and no need to evaluate the rest.

However, this comment gives me the thought that iterating over the commits is extraneous. I used the `small_patch.py` plugin as a template to munge, which lead me to iterate over commits. My original prototype simply iterated over the file 
info returned via a base...head comparison (`pygithub.repo.compare()`). I could ostensibly use `issuewrapper.get_files()` to get only the file info we use here.

ansibullbot/triagers/plugins/docs_info.py
sommersoft: Bad habit to shortcut the desired state and avoid a couple lines of code. Having it failsafe to `False` is an excellent point.

ansibullbot/triagers/plugins/docs_info.py
mkrizek: > This is to break out of the parent loop that is iterating over each commit. This is basically implementing a "fail fast" scenario, since there was a non-docs change and no need to evaluate the rest.

I guess my point was that _check_py_changes is called per changed_file but nothing is done with the result. Shouldn't we break out of the "per changed_file" loop as soon as _check_py_changes returns False?

> However, this comment gives me the thought that iterating over the commits is extraneous. I used the small_patch.py plugin as a template to munge, which lead me to iterate over commits. My original prototype simply iterated over the file
info returned via a base...head comparison (pygithub.repo.compare()). I could ostensibly use issuewrapper.get_files() to get only the file info we use here.

+1

acozine:
Thanks @sommersoft for a detailed PR! This sent me back to look at the labels more generally.

We have three "docs-related" labels: `docs`, `docsite`, and `docsite_pr` (see [descriptions](https://github.com/ansible/ansible/labels?page=12&sort=name-asc)). The `docsite_pr` label is meant for PRs that use the `Edit on GitHub` feature. I'm not sure that label is being applied correctly, but the intention is clear and, if correctly applied, useful. The other two labels are less clearly differentiated in my mind, and the descriptions don't help very much. It would be great to make the descriptions and the logic clear. 

If I'm reading this PR correctly, it would use `docsite` for PRs and issues that ONLY affect the docs, and `docs` for PRs and issues that include both docs changes and other changes. That seems reasonable, but a broader discussion would be worthwhile, so we can update the descriptions as well as the code. Can we add this to the [DaWGs agenda](https://github.com/ansible/community/issues/579) for next week?

sommersoft:
> If I'm reading this PR correctly, it would use `docsite` for PRs and issues that ONLY affect the docs, and `docs` for PRs and issues that include both docs changes and other changes.

As it is now, this PR will only apply `docsite` for PRs that only affect the docs. Issues are not evaluated, and `docs` is not applied in any case.

> We have three "docs-related" labels: docs, docsite, and docsite_pr (see descriptions). The docsite_pr label is meant for PRs that use the Edit on GitHub feature. I'm not sure that label is being applied correctly, but the intention is clear and, if correctly applied, useful. The other two labels are less clearly differentiated in my mind, and the descriptions don't help very much. It would be great to make the descriptions and the logic clear.

As @mkrizek pointed out and I agree with, a new `docs_only` label might work better for what this PR is doing. However, if one of the current label descriptions is defined to capture this meaning, using one of them could work too. This is all about making things easier for the humans, after all.

> Can we add this to the DaWGs agenda for next week?

Certainly.

ansibullbot/triagers/plugins/docs_info.py
sommersoft: Using `issuewrappper.get_files()` (broken, btw) and/or `issuewrapper.files()` doesn't work how I assumed. They only return the filename and the sha, from the meta info. I'd likely have to build out a new solution to get something like `compare base...head`, for the relevant API.

Attempting this though did highlight that the purpose of iterating over commits may be so that the meta history can be built.

I'll stick with iterating over the commits for now, but will work to get a better understanding of how that affects the desired outcome.

acozine:
Added a label called `docs_only`. Per discussion in the DaWGs meeting today, we will keep the `docs` label as a generic one for all issues and PRs that relate to documentation (including issues created with the docs issue template and PRs that involve code changes as well as documentation changes). We'll add the `docsite_pr` label to PRs created using `Edit on GitHub` feature. And we'll add `docs_only` for PRs that have only documentation changes.

In time, we may be able to deprecate and/or delete the `docsite` label, but we won't try that until the other three are working well.

acozine:
Thanks @sommersoft for updating this to use `docs_only`.

@mkrizek are you comfortable merging this? Docs team is +1 on the functionality, but I can't vouch for the code.

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+
+    for commit in iw.commits:
+        commit_files = iw.get_commit_files(commit)
+        if commit_files is None:
+            # "Sorry, this diff is temporarily unavailable due to heavy server load."
+            # Preserve docsite label to prevent potential waffling
+            dfacts["is_docs_only"] = "docs_only" in iw.labels
+            return dfacts
+
+        has_non_docs_changes = any(filterfalse(_is_docs_only, commit_files))
```
mkrizek, line 298: I think this is too clever and the readability would be better with classic `for` loop here - the combination of `has_non_docs_changes`, `any`, `filterfalse` and `_is_docs_only` confused me at first. Just a note, I think I am fine with staying this as is though.

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+
+def _is_docs_only(changed_file):
+    """ Check if the changes made to ``changed_file`` affect only documentation. """
+    docs_only = False
+
+    if isinstance(changed_file, dict):
+        changed_file = CommitFile(changed_file)
+
+    if _is_docs_path(changed_file.filename):
+        docs_only = True
```
mkrizek, line 259: I think if we just `return` here and do the same for the rest of the function we can get rid of all `else`s and reduce the complexity here.

mkrizek: So we would end up with:
```python
def _is_docs_only(changed_file):
    """ Check if the changes made to ``changed_file`` affect only documentation. """
    if isinstance(changed_file, dict):
        changed_file = CommitFile(changed_file)

    if _is_docs_path(changed_file.filename):
        return True

    # Additions or deletions of complete files outside of
    # docsite/examples will likely be more than documentation
    # changes
    if changed_file.status != "modified":
        return False

    # Non-Python files that are outside of the docsite
    # folder will not be documentation related
    if not changed_file.filename.endswith(".py"):
        return False

    # If a python file, check if the changes are only to
    # docstrings
    if _check_py_changes(changed_file.file_content, changed_file.patch):
        return True

    return False
```

mkrizek: Hrm, and we can go further and just do:
```python
def _is_docs_only(changed_file):
    """ Check if the changes made to ``changed_file`` affect only documentation. """
    if isinstance(changed_file, dict):
        changed_file = CommitFile(changed_file)

    return (
        _is_docs_path(changed_file.filename) or _check_py_changes(changed_file.file_content, changed_file.patch)
    )
```
?

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+                            func_map.ds_line_end = (
+                                func_map.ds_line_start + len(func_doc.splitlines())
+                        )
+                    class_map.funcs.append(func_map)
+
+            mod_map.classes.append(class_map)
+
+    return mod_map
+
+def _check_py_changes(file_content, diff):
```
mkrizek, line 211: This function should be renamed to reflect the intent.

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+    def file_content(self):
+        if self.raw_url:
+            result = requests.get(self.raw_url)
+            if result.ok:
+                return result.text
+
+def _is_docs_path(filename):
+    """ Determine if affected file is only applicable to documentation directories """
+    for pattern in RE_DOCS_PATTERNS:
+        if re.search(pattern, filename):
```
mkrizek, line 89: Should we use here just `.startswith` instead of regex?

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
...
+        changes.
+    """
+
+    diff = _get_diff_info(diff)
+    source = None
+
+    if file_content is not None:
+        source = _get_ast_info(file_content)
+
+    if source is not None:
```
mkrizek, line 222: We could just invert the conditional and return early here.

ansibullbot/triagers/ansible.py
```diff
@@ -1285,6 +1286,18 @@ def create_actions(self, iw, actions):
...
+        # docs_only?
+        # https://github.com/ansible/ansibullbot/issues/1047
+        if iw.is_pullrequest():
+            label_name = 'docs_only'
+            if self.meta['is_docs_only']:
+                if label_name not in iw.labels:
+                    actions.newlabel.append(label_name)
+            else:
+                human_labeled = iw.history.was_labeled(label_name, bots=self.BOTNAMES)
+                if label_name in iw.labels and not human_labeled:
```
mkrizek, line 1298: With this we could end up with the following undesired situation:
1. a human adds `docs_only` label
2. a submitter adds a new commit that changes code -> `is_docs_only` is set to `False`
3. ansibot does not remove `docs_only` label because a human added `docs_only` label in the past.

If we want to include the "human_labaled" check we need to be more careful about it:
* was it unlabeled by human afterwards?
* was there a new commit after human added the label?

It might be just easier to not take `human_labeled` into account but I am not against implementing it.

ansibullbot/triagers/plugins/docs_info.py
mkrizek: And we should `return False` if source is `None` instead of returning `None` (implicitly) in that case.

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,306 @@
+import ast
+import dataclasses
+from itertools import filterfalse
+import logging
+import re
+
+import requests
+
+RE_DOCS_PATTERNS = [
```
mkrizek, line 9: @acozine Should we include `lib/ansible/plugins/doc_fragments/` as well?

mkrizek: Also, if bot determines that a PR is `docs_only` and only other changes in addition to docs only are in `changelogs/fragments`, we might add the label anyway?

mkrizek (changes requested):
Thanks for adding the tests @sommersoft!

Overall it looks good. I added comments but most of them are just suggestions, there are a few that need to be addressed though.

ansibullbot/triagers/plugins/docs_info.py
sommersoft: I have had the same thoughts, while writing it and have been stewing on it. I'll work some more iterations to improve readability.

sommersoft: So two things on this:

1. I realize now that I had tunnel vision on finding falsey cases, which lead to the cleverness.
2. I re-visited `issuewrapper.get_files()` from https://github.com/ansible/ansibullbot/pull/1505#discussion_r574552256. When I looked at it before, I made my conclusion based only on the repr. It actually has everything we need, and would allow to only focus on the changed content and not each commit. I'm still not sure if the botmeta desires the by-commit-history, though. With `get_files()` and can actually get this down to one or two lines:
   ```python
    pr_files = iw.get_files()
    docs_only = all([_is_docs_only(f.raw_data) for f in pr_files])

    # OR (without all(); explicitly looking for False)
    pr_files = iw.get_files()
    docs_only = False not in [_is_docs_only(f.raw_data) for f in pr_files]
   ```
   This would drive a fix to `get_files()`, as it currently tries to set a non-existent attribute (`.files`). Not sure if it should work _with_ the `files` property, or separately. Would also drive more changes to `IssueMock` for testing, but that is trivial.

ansibullbot/triagers/plugins/docs_info.py
sommersoft: Using `.startswith` is fine by me. The regex approach is just a pre-optimization for cases that are more complex than just the root path.

ansibullbot/triagers/plugins/docs_info.py
mkrizek: > I'm still not sure if the botmeta desires the by-commit-history, though.

Can you elaborate on the above? What do you mean by `botmeta`?

sommersoft: > Can you elaborate on the above? What do you mean by botmeta?

This derived from my uncertainty in how the meta/history worked with cached info. I was afraid that it _needed_ the history to be ordered by commits.

I took some more time to study the botmeta & history caching. I understand now that it is all cached by the issue/PR, so using the commits isn't necessarily required.

ansibullbot/triagers/plugins/docs_info.py
sommersoft: I went with the longer form, to preserve the intent comments.

ansibullbot/triagers/ansible.py
sommersoft: Doh! I forgot about this one. I can take it out and just follow the norm of removing as the others do.

sommersoft:
I've updated to using `get_files()` and all associated logic. The test changes weren't as trivial as I expected, but aren't too major.

I apologize for adding to the review workload on this!

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,294 @@
...
+    dfacts = {
+        "is_docs_only": False
+    }
+
+    if not iw.is_pullrequest():
+        return dfacts
+
+    docs_only = False
+
+    pr_files = iw.pullrequest.get_files()
```
mkrizek, line 290: Please us `iw.files` at it uses cache.

ansibullbot/triagers/plugins/docs_info.py
```diff
@@ -0,0 +1,294 @@
...
+def get_docs_facts(iw):
+    """ Cycle through the files and gather facts about documentation changes. """
+    dfacts = {
+        "is_docs_only": False
+    }
+
+    if not iw.is_pullrequest():
+        return dfacts
+
+    docs_only = False
```
mkrizek, line 288: We can remove this.

ansibullbot/triagers/plugins/docs_info.py
sommersoft: `iw.files`, as it is currently written, only returns a list of the filenames. It does not return the full list of `self.pr_files`, with each item containing the attributes we subsequently use.

My previous changes in https://github.com/ansible/ansibullbot/pull/1505/commits/19e77980360c9bc982c6a70c605434267d36730a utilized the cached approach, while leaving `iw.files` as-is.

mkrizek: Ah, didn't realize that. Then using `iw.load_update_fetch_files()` is what we want here.

And sorry for changing those methods while you were working on this PR.

sommersoft: > And sorry for changing those methods while you were working on this PR.

No need to be sorry. The cleanup was a good thing, and `get_files` was broken and unused.

> Then using `iw.load_update_fetch_files()` is what we want here.

I'll work towards using `load_update_fetch_files` directly. The mocking in `IssueMock` is doable, but could be confusing for future uses as it would be completely different in functionality.

Do we maybe want to re-introduce `get_files`? Or, as my brain over-complicates things, make `self.pr_files` a property:

```py
class DefaultWrapper:
    def __init__(self, ...):
        #self.pr_files = None -->
        self._pr_files = None

    @property
    def pr_files(self):
        if self._pr_files is None:
            self._pr_files = self.load_update_fetch('files')
        return self._pr_files

    @property
    def files(self):
        if self.is_issue():
            return None
        return [x.filename for x in self.pr_files]
```

mkrizek: I am fine with the above addition.

mkrizek (approved):
Thanks!

mkrizek:
@sommersoft I see there is `WIP` in title, was there anything else you wanted to add here?

sommersoft:
> I see there is WIP in title, was there anything else you wanted to add here?

I don't have anything else. Still awaiting an answer from acozine on your question of additional exclusions, but I don't think that should necessarily hold this up.

PR #1549, opened 2021-05-05
assertEquals is deprecated

mkrizek:


webknjaz:
FWIW I'd recommend switching over to `assert x == y` syntax as it's what works best with pytest.

mkrizek:
Yeah, that is the end goal, to migrate away from `unittest` (and increase code coverage, make unit tests faster, ...).

webknjaz:
Yep, OTOH this specific change doesn't require removing `import unittest` and can be done right away.

PR #1583, opened 2021-06-15
triagers/ansible: un-hardcode ansible_core_team

dmsimard:
Maintainers and committers may be different across different
repositories. As such, we can't hardcode the organization name "ansible"
or the teams so we're pulling it back into a configuration parameter instead.

Since the maintainers are not exclusively ansible core maintainers,
we're also renaming the property to something a bit more generic.

### DIFF
--- a/ansibullbot/triagers/defaulttriager.py
+++ b/ansibullbot/triagers/defaulttriager.py
@@ -118,6 +118,20 @@ class DefaultTriager:
             server=C.DEFAULT_GITHUB_URL
         )
 
+        self._maintainer_team = None
+
+    @property
+    def maintainer_team(self):
+        # Note: this assumes that the token used by the bot has access to check
+        # team privileges across potentially more than one organization
+        if self._maintainer_team is None:
+            self._maintainer_team = []
+            teams = C.DEFAULT_GITHUB_MAINTAINERS
+            for team in teams:
+                _org, _team = team.split('/')
+                self._maintainer_team.extend(self.ghw.get_members(_org, _team))
+        return list(set(sorted([x for x in self._maintainer_team if x not in C.DEFAULT_BOT_NAMES])))
+
     @classmethod
     def create_parser(cls):
         parser = argparse.ArgumentParser()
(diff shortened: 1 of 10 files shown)


ansibullbot/triagers/defaulttriager.py
```diff
@@ -118,6 +118,20 @@ def __init__(self, args=None):
...
+    def maintainer_team(self):
+        # Note: this assumes that the token used by the bot has access to check
+        # team privileges across potentially more than one organization
+        if self._maintainer_team is None:
+            self._maintainer_team = []
+            teams = C.DEFAULT_GITHUB_MAINTAINERS
+            for team in teams:
+                _org, _team = team.split('/')
+                self._maintainer_team.extend(self.ghw.get_members(_org, _team))
+        return list(set(sorted([x for x in self._maintainer_team if x not in C.DEFAULT_BOT_NAMES])))
```
mkrizek, line 133: I would move the `list(set(sorted...` processing to the above `if` branch as well.

Also I am curious why passing `sorted` into `set` since `list` makes it unsorted again? Would the following make more sense?
```python
sorted(set([x for x in self._maintainer_team if x not in C.DEFAULT_BOT_NAMES]))
```

Or just do:
```python
sorted(set(self._maintainer_team).difference(C.DEFAULT_BOT_NAMES)))
```

ansibullbot/triagers/defaulttriager.py
```diff
@@ -130,7 +130,7 @@ def maintainer_team(self):
             for team in teams:
                 _org, _team = team.split('/')
                 self._maintainer_team.extend(self.ghw.get_members(_org, _team))
-        return list(set(sorted([x for x in self._maintainer_team if x not in C.DEFAULT_BOT_NAMES])))
+        return sorted(set(self._maintainer_team).difference(C.DEFAULT_BOT_NAMES)))
```
dmsimard, line 133: I tested it and it works fine:
```
before: ['Ompragash', 'dericcrago', 'gundalow', 'dmsimard', 'tadeboro', 'felixfontein', 'abadger', 'Andersson007']
after: ['Andersson007', 'Ompragash', 'abadger', 'dericcrago', 'dmsimard', 'felixfontein', 'gundalow', 'tadeboro']
```

